Issue #33, opened 2023-10-03
React issues 


kylebarron:
Goals:

- Render multiple dataframes/layers on a single map
- Enable updates of Python properties like `get_fill_color` to propagate to the map

#### Python side

On the Python side I attempted to write a sort of "container widget" (ref https://github.com/manzt/anywidget/pull/194), where each `Layer` object is a Python jupyter widget, and where the `Map` object collects each of the underlying Widgets. It's useful to have each Layer be its own Widget, because that enables event handling on each layer.

```py
from ipywidgets import Widget # Base widget class
from anywidget import AnyWidget # high level widget helper that subclasses Widget

class BaseLayer(Widget):
    """Base class for our layer types"""
    ...

class PointLayer(BaseLayer):
    ...

class LineStringLayer(BaseLayer):
    ...

class PolygonLayer(BaseLayer):
    ...

class Map(AnyWidget):
    _esm = "path to esm JS bundle"
    _css = "optional path to CSS styling"

    # list of instances of classes that subclass from BaseLayer
    layers = List[BaseLayer]
```

Then a user will create a variety of layers and instantiate a `Map` object:

```py
import geopandas as gpd

point_data = gpd.GeoDataFrame(...)
polygon_data = gpd.GeoDataFrame(...)

point_layer = PointLayer(point_data)
polygon_layer = PolygonLayer(polygon_data)

map = Map(layers=[point_layer, polygon_layer])
map # putting this last in a cell "prints" the object..., which in this case renders the map
```

The goal with this setup is to let a user run

```py
point_layer.fill_color = [255, 0, 0]
```
and the points on the map turn red.

In order for this to happen, the JS side needs to be able to receive these events and re-render

#### JS Side

When you render the map object

This will then sync data with the JS side and render the `App` object
https://github.com/developmentseed/lonboard/blob/8cd1a19796dd529000c33c9da99ec2532d7ed499/src/index.tsx#L142

[...]

PR #34, opened 2023-10-03
Support multiple layers on a single map

kylebarron:
### Change list

- Support for rendering multiple layers on a single map.
- The serialization on the top-level model only passes _string keys_ to the frontend. Then to access the actual sub-model containing each layer, the frontend calls the async `WidgetManager.get_model()`. In this PR's implementation, it manages a `subModelState` object that contains a mapping from layer id to model data. Each model data now has a class (like `ScatterplotModel`) that wraps each layer model with strong typing and which also manages instantiating event handlers to update the map when the model data changes.
- 

### todo:

Follow ups for future prs:

- [ ] Remove support for rendering layers in isolation (outside of a map). Or maybe overwrite the `__repr__` to make the repr of a layer pass its data into `Map()`? Thus making it easy to render just one layer, but directing people to create a `Map` directly for more control. 
- [ ] No longer subclass layers from AnyWidget
- [ ] Update inference of `initial_view_state` for the top-level `Map` object
- [ ] Remove `.tsx` apps for specific layers

Closes https://github.com/developmentseed/lonboard/issues/33


package.json
```diff
@@ -14,6 +14,7 @@
     "react-map-gl": "^7.1.5"
   },
   "devDependencies": {
+    "@jupyter-widgets/base": "^6.0.6",
```
kylebarron, line 17: This is a type-only dependency, so I think it's ok to have it as a dev dependency. And probably better this way, as it won't interact with whatever version jupyterlab is using.

src/index.tsx
```diff
@@ -21,163 +25,118 @@ const INITIAL_VIEW_STATE = {
...
+  return await Promise.all(promises);
 }
 
 function App() {
+  let [parquetWasmReady] = useParquetWasm();
+  let [initialViewState] = useModelState<DataView>("_initial_view_state");
+
+  let [subModelState, setSubModelState] = useState<
+    Record<string, BaseGeoArrowModel>
+  >({});
```
kylebarron, line 50: This manually stores the state of all child layers

kylebarron:
This is working now!
Screenshot 2023-10-30 at 5 56 34 PM


Works like a charm in colab too:
image

kylebarron:
@vgeorge I think this is ready to merge if you'd like to take a look at it! Any feedback is appreciated! It looks like it works, but maybe there's a cleaner way to implement it 🤷‍♂️

kylebarron:
Merging now to make follow up prs off the main branch, but feedback still welcome

PR #76, opened 2023-10-11
Fix map attribution

### LINKED ISSUE #13: OSM attribution in corner of map
The current map is missing any attribution, and should be fixed

kylebarron:
Really no idea why this isn't working. None of the deck.gl examples are up to date with the latest react-map-gl 😕 

Ref:

- [`AttributionControl` docs](https://visgl.github.io/react-map-gl/docs/api-reference/attribution-control)

Closes #13


vgeorge:
@kylebarron the attribution issue seems to be related to the missing Maplibre CSS styles file. When inspecting the page, it is possible to see the attribution elements in the DOM, but no styles are applied to them and they end up positioned outside the widget viewport. I tried to import the CSS like instructed in [the docs](https://visgl.github.io/react-map-gl/docs/get-started), but it didn't work. I'll keep looking.

kylebarron:
@vgeorge it turns out that it's related to the heights of the multiple divs. So if I change to

```diff
      <DeckGL
        initialViewState={
          ["longitude", "latitude", "zoom"].every((key) =>
            Object.keys(viewState).includes(key)
          )
            ? viewState
            : DEFAULT_INITIAL_VIEW_STATE
        }
        controller={true}
        layers={layers}
+       height={300}
        // ContextProvider={MapContext.Provider}
      >
```
Then the map inside the main div is smaller and we can see the attribution below it (but weirdly in a drop down)
image

vgeorge:
@kylebarron I was able to see the attribution elements after following your instruction of adding the height to the DeckGL component. But the Maplibre CSS module is not imported, it seems to me the dropdown is a result of missing CSS styles. Maybe something wrong in build.mjs?

kylebarron:
> But the Maplibre CSS module is not imported

I'm not good at checking this from the browser console; https://github.com/developmentseed/lonboard/pull/76/commits/17dc0376fa08a62480dea8ae07848cac244f75b5#diff-c2f8ec75103ae661341f239c22352142d1d5e965cad55b19cfb978902500106cR9 didn't fix it? 

Alternatively we could build our own CSS bundle that we dynamically load through the ESM endpoint. You can pass css either inline as a string or as a url in `_css` in the anywidget object: https://anywidget.dev/en/getting-started/

kylebarron:
Thanks @vgeorge for your help! 🙏

kylebarron:
The map attribution still isn't _clickable_, but that's a separate longstanding deck/react-map-gl issue that I don't know what the current solution is https://github.com/visgl/deck.gl/issues/4165

PR #100, opened 2023-10-13
fix release

kylebarron:
The current release is broken because the JS files are never built.

### Change list

- Build js files in CI
- Check that files exist in the wheel with a path of `lonboard/static`


kylebarron:
Ref https://github.com/manzt/anywidget/issues/344

PR #132, opened 2023-10-17
Default to excluding pandas index

kylebarron:


kylebarron:
hmm might have some issues around empty tables...

Issue #36, opened 2023-10-03
Tooltip


kylebarron:
It would be great to have a tooltip that shows the row of data when hovered or clicked. This relies on https://github.com/geoarrow/deck.gl-layers/issues/30

kylebarron:
Closed by https://github.com/developmentseed/lonboard/pull/147 and https://github.com/developmentseed/lonboard/pull/151

PR #147, opened 2023-10-25
Add tooltip

vgeorge:
Contributes to #36. This is a work in progress.

Changes:

- Enabled tooltip on hover for all layers
- Defined a common getTooltip() function that generates a formatted table with the feature properties
- Modified the build watcher to observe `.css` files

Issues:

- Feature properties don't appear to be available in `info.object`. Should we deserialized them from the arrow object?
- A small black square appears beneath the tooltip

cc @kylebarron


kylebarron:
My changes:

- Set [pickingRadius](https://deck.gl/docs/api-reference/core/deck#pickingradius) to 10 pixels to make it easier to pick small objects
- Return only the _content_ of the html div from `getTooltip`

PR #151, opened 2023-10-27
Fix tooltip styles

vgeorge:
Fix some issues of #147. Changes:

- Move tooltip wrapper styles from CSS file to `style` property. They aren't taking effect because of inline styles
- Make table border and property id font lighter

@kylebarron ready for review.


src/tooltip/index.ts
```diff
@@ -36,6 +36,12 @@ export function getTooltip({ object }: GeoArrowPickingInfo): TooltipContent {
     return {
       className: "lonboard-tooltip",
       html: toHtmlTable(jsonObj),
+      style: {
+        backgroundColor: "#fff",
+        boxShadow: "0 0 15px rgba(0, 0, 0, 0.1)",
+        color: "#000",
+        padding: "10px",
```
kylebarron, line 43: It looks a little cleaner with a smaller padding in my own opinion 🤷‍♂️ .
```suggestion
        padding: "6px",
```

kylebarron (approved):
Thanks!

PR #171, opened 2023-10-31
Try to add versioned docs

kylebarron:
This is kinda tough to get working, and without testing in prod 🤷‍♂️ 

When I run this locally, it ignores my gitignore and tries to push all of `node_modules`, `.cache`, etc to the branch


kylebarron:
Appears to work 🤷‍♂️  
image

PR #193, opened 2023-11-01
Layer extensions

kylebarron:


kylebarron:
Todo: check what happens when multiple layers use the same extension, as in the arc-brushing example

kylebarron:
This will probably stay in "experimental" for a bit, but good enough to merge I thinkk

PR #218, opened 2023-11-07
Support `pyarrow.Table` with `geoarrow.pyarrow` extension types as geometry columns

jorisvandenbossche:
Work-in-progress, and haven't yet tested it.

The idea is that when encountering a `geoarrow.pyarrow` extension type, to convert this to its raw storage field with metadata (i.e. how the extension type looks like in IPC or when the python wrapper class is not registered). Short term, this is the easiest approach, because a lot of the existing code relies on checking eg the "ARROW:extension:name" in the field metadata.


jorisvandenbossche:
This works now, but it still requires you to ensure you are passing interleaved coordinates, and not one of the other geoarrow types (separated, or wkb/wkt). We could also do that conversion automatically.

Should probably factor out the added code into a helper function somewhere.

kylebarron:
> This works now, but it still requires you to ensure you are passing interleaved coordinates, and not one of the other geoarrow types (separated, or wkb/wkt). We could also do that conversion automatically.
> 
> Should probably factor out the added code into a helper function somewhere.

For now I'd be happy to just assert that the coordinates are interleaved, and handle separated/wkt/wkb in the future

kylebarron (approved):

kylebarron:
this lgtm! Anything else you wanted to do before merging?

kylebarron:
In #308 I now check for pycapsule interface methods but don't yet handle pyarrow extension types

Issue #167, opened 2023-10-30
Let widget fill available height


kylebarron:
When using with [jupyter sidecar](https://github.com/jupyter-widgets/jupyterlab-sidecar), the map can be placed on the right side of the notebook screen:

<img width="980" alt="image" src="https://github.com/developmentseed/lonboard/assets/15164633/122ba2c4-3915-482b-aaab-4df4cf58676d">

Right now the div containing the deck.gl widget is hard-coded to 500px:
https://github.com/developmentseed/lonboard/blob/d4a05a3a2f3458eae1aba41ca879d1d40886f068/src/scatterplot-layer.tsx#L80

This means that when used with sidecar, if the screen is more than 500px tall, it'll have a weird empty space at the bottom. 

Ideally we want to let the widget fill all available height in its containing div, but I can't figure out how to do that. When I switch to, say,
```tsx
style={{ display: "flex", flexFlow: "column", flexGrow: 1, overflow: "auto" }}
```
it creates a div with zero height:
<img width="1269" alt="image" src="https://github.com/developmentseed/lonboard/assets/15164633/478789c1-b326-4ef1-a2c1-ea19471c0438">

cc @vgeorge

kylebarron:
Closed by https://github.com/developmentseed/lonboard/pull/220

kylebarron:
HTML map height is tracked in https://github.com/developmentseed/lonboard/issues/247

PR #220, opened 2023-11-08
Improve map height

vgeorge:
Contributes to #167, WIP.

With this change, if the user doesn't set `_height` to Map, the component will try to dynamically occupy the full height of the parent element. 

This will make the map height when using Sidecar to occupy the full height, which is nice, but it will require the user to know that  `_height` must be set when not using sidecar, otherwise, the widget will look like this:

Screenshot 2023-11-08 at 16 42 06

I believe we need to figure out how to set a default `minHeight` so the map would display well on both use cases.

cc @kylebarron


vgeorge:
@kylebarron I added `min-height` to the widget wrapper and it seems to be working, could you please review?

kylebarron:
This doesn't seem to be working for me without sidecar in JupyterLab...
image

After I manually set the height it works again
image

src/index.tsx
```diff
@@ -151,8 +151,29 @@ function App() {
...
+        // Set the height to 100% and min-height
+        mapContainerParent.style.height = "100%";
+        mapContainerParent.style.minHeight = "500px";
+      }
+    }
+  }, []);
+
   return (
-    <div style={{ height: mapHeight || 500 }}>
+    <div id="map" style={{ height: mapHeight || "100%" }}>
```
kylebarron, line 176: I don't think `id="map"` is sufficient because multiple views (of the same or different widgets) could be shown in the same notebook. (This could also be my rendering issue if it works fine for you)

kylebarron:
I'm not sure if the latest change to have a single uuid div per map works either... Ideally we want to have a unique id per _view_ not a unique id per _model_. For example, it still breaks if I first render a map inline and _then_ try to move the map to a sidecar:

image

notebook code:

```
import geopandas as gpd
from lonboard import viz
from sidecar import Sidecar

gdf = gpd.read_file(gpd.datasets.get_path('nybb'))
map_ = viz(gdf)
map_

sidecar = Sidecar()
with sidecar:
    display(map_)
```

presumably it's finding the _wrong_ div

kylebarron:
Oh smart, that looks great! Thanks!

PR #226, opened 2023-11-10
Add docs for installation from repository

emileten:
If installing from source, one needs to manually build the JS components. 

See https://github.com/developmentseed/lonboard/discussions/225

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,13 @@ Python library for fast, interactive geospatial vector data visualization in Jup
 pip install lonboard
 ```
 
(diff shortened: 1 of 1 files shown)


kylebarron:
I'd prefer to keep the instructions in the main `README.md` as minimal as possible, because there will be a lot of less-technical people reading that, and it's also the front page for the website: https://developmentseed.org/lonboard/latest/.

How about:

> To install from source, follow the [contributor documentation](https://github.com/developmentseed/lonboard/blob/main/DEVELOP.md).

And then we can improve the docs in that page if needed

PR #234, opened 2023-11-10
Enable watchfiles

kylebarron:


kylebarron:
Superseded by #235 , though the root problem is unresolved

PR #251, opened 2023-11-15
Add method to apply a categorical colormap

kylebarron:
### Change list

- Add a function `apply_categorical_cmap` that applies colors from a categorical column.

This is 180-200x faster than leafmap's existing `assign_discrete_colors` method on string data. When the user already has their data stored as a categorical, it's 620x faster than leafmap (although leafmap seemed to crash on the categorical I tested, so that's comparing `apply_categorical_cmap` on the categorical with `assign_discrete_colors` on the string). 

On the [dc wetlands dataset](https://data.source.coop/giswqs/nwi/wetlands/DC_Wetlands.parquet) with 1500 rows:

image

Expanding the rows 1000 times to test with 1.5 million rows:

image

And when the input column is already a pandas categorical, this drops to 25ms.

image

### DIFF
--- a/docs/api/colormap.md
+++ b/docs/api/colormap.md
@@ -1,3 +1,9 @@
 # lonboard.colormap
 
 ::: lonboard.colormap.apply_continuous_cmap
+
+::: lonboard.colormap.RGBColor
+
+::: lonboard.colormap.DiscreteColormap
+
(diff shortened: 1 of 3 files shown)


giswqs:
Nice! I will switch to use `apply_categorical_cmap` in leafmap.

PR #257, opened 2023-11-21
Enable TS strict mode

kylebarron:
Just a stub to come back and ensure we're meeting ts strict mode


kylebarron:
Closed in favor of https://github.com/developmentseed/lonboard/pull/271

Issue #247, opened 2023-11-13
Improve map height for static HTML export


kylebarron:
In #220 @vgeorge was super helpful in fixing the map height inside JupyterLab when using sidecar. If I remember correctly, it relied on one of the components above it having absolute positioning or something like that.

There's a `to_html` method on the map which generates a static HTML page with the map and data embedded. But when I save output from a widget, the height is still limited to 500 pixels. 

<img width="1511" alt="image" src="https://github.com/developmentseed/lonboard/assets/15164633/783d4f51-cb63-4e1f-a6b8-a53f5471b16d">

If it does have something to do with the structure of the generated HTML, we might want to use our own HTML template instead of relying on the default one from `ipywidgets`. See https://ipywidgets.readthedocs.io/en/latest/embedding.html#python-interface for docs on this.

PR #283, opened 2023-12-02
Set map's parent display to flex

jtmiclat:
## What I am changing
Fixing the default height problem 

## How I did it
Main factor here is setting the map's the parent's div's display to flex.

The default `display:block` will try to cover the entire the contents if set to 100% but since at initializing there is no contents/canvas/children for the parent this would be 0px. While `display:flex` bypasses this.  

Another common solution for web app is to set the height and width to `100vh` and `100vw` but this assumes that the map will cover the entire viewport/iframe/window which might not work for jupyter widgets, sidepanels.

## How you can test it
Export a map using the following code:

```python
import geopandas as gpd
from lonboard import viz
gdf = gpd.read_parquet(...) # any file would do
map = viz(gdf)
map.to_html("test.html")
```

Before: 
Screenshot 2023-12-02 at 6 14 33 PM

After:
Screenshot 2023-12-02 at 6 13 37 PM

Let me know if you have other suggestions on how to tests this! 

## Related Issues
Related to #247

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -118,9 +118,10 @@ function App() {
 
       // Check if the height is not already set
       if (!parentStyle.height || parentStyle.height === "0px") {
-        // Set the height to 100% and min-height
+        // Set the height to 100%, min-height, and display to flex
         mapContainerParent.style.height = "100%";
         mapContainerParent.style.minHeight = "500px";
+        mapContainerParent.style.display = 'flex';
       }
     }
   }, []);


kylebarron:
Hmm I went to your branch and tried this... 
image

But it still doesn't appear to fill the height
image

jtmiclat:
Sorry, I think I screwed up and made a mistake with my assumption and tests. Closing for now until I can figure out how to address it!

Issue #248, opened 2023-11-13
Fix CSS in static HTML export


kylebarron:
Given default fonts, presumably css is not getting added to the static HTML page

<img width="840" alt="image" src="https://github.com/developmentseed/lonboard/assets/15164633/573a3be2-a686-482c-8160-3609c9392d76">

giswqs:
This issue might be relevant: https://github.com/opengeos/leafmap/issues/594#issuecomment-1807013768

kylebarron:
I don't think that's related. The standalone html export is working fine here; it's just missing our css.

giswqs:
Never mind. The exported html doesn't work with Streamlit. I will look into it later.

kiasari:
@giswqs Could you find a way to display the `Map` object in Streamlit?

PR #284, opened 2023-12-02
Add font to index.css

jtmiclat:
## What I am changing
Add missing css to lonboard table

## How I did it
The problem is that when inside a jupyter notebook, the tooltip (and other html elements) inherit the some css from jupyter lab/notebook compared to running it in exported form. In this cases, it was the font. 

Here is the css of the tooltip when in jupyter lab. The one declaring the font is not related to the css of lonboard but the css of the body.
Screenshot 2023-12-02 at 8 25 56 PM

vs css of the tooltip when exported
Screenshot 2023-12-02 at 8 30 27 PM

To fix this I added the missing font to the lonboard css. 

## How you can test it
Export map
```python
import geopandas as gpd
from lonboard import viz

gdf = gpd.read_parquet(...) # any file will do
map = viz(gdf)
map.to_html("test.html")
```

Before 
Screenshot 2023-12-02 at 8 33 22 PM

After
Screenshot 2023-12-02 at 8 34 02 PM

## Related Issues
#248

### DIFF
--- a/src/tooltip/index.css
+++ b/src/tooltip/index.css
@@ -1,4 +1,9 @@
 /* Remove spaces between table cells */
+.lonboard-tooltip {
+  font-family: var(--jp-ui-font-family);
+  font-size: var(--jp-ui-font-size1);
+}
+
 .lonboard-tooltip table {
   border-collapse: collapse;
 }


kylebarron:
Thanks! Looks good; I'll do some quick manual tests on Monday and then merge

kylebarron:
Looks great! Thanks!

Issue #206, opened 2023-11-06
BitmapTileLayer


kylebarron:
Auth questions:

- For now should this only work with public data?
- How do we want to support non-public data in the future?

PR #285, opened 2023-12-02
[WIP] Bitmap and Bitmap Tile Layers

jtmiclat:
Addresses #206 
Example usage:
```python
import geopandas as gpd
from lonboard import Map
from lonboard._layer import  BitmapLayer, BitmapTileLayer


layer = BitmapLayer(bounds=[-122.5190, 37.7045, -122.355, 37.829],image='https://raw.githubusercontent.com/visgl/deck.gl-data/master/website/sf-districts.png')
map_ = Map(layers=[layer])
map_
```
Screenshot 2023-12-03 at 5 23 37 AM

and 
```python
layer = BitmapTileLayer(data='https://c.tile.openstreetmap.org/{z}/{x}/{y}.png')
map_ = Map(layers=[layer], _initial_view_state={'longitude': -122.42060000000001, 'latitude': 37.7543, 'zoom': 11})
map_
```

Screenshot 2023-12-03 at 5 23 26 AM

Decision Points
1. Best way to handle non arrow layers. Should we have new classes for non arrow layers that dont include fields like `table` or should we be extending existing classes and setting table to `null`/`None`
 2. What parameters should we include for both layers?

### DIFF
--- a/lonboard/_layer.py
+++ b/lonboard/_layer.py
@@ -7,8 +7,8 @@ from typing import TYPE_CHECKING
 import geopandas as gpd
 import ipywidgets
 import pyarrow as pa
-from shapely.geometry import box
 import traitlets
+from shapely.geometry import box
 
 from lonboard._base import BaseExtension, BaseWidget
 from lonboard._constants import EPSG_4326, EXTENSION_NAME, OGC_84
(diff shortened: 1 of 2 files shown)


kylebarron:
This looks awesome! Thanks!

> 1. Best way to handle non arrow layers. Should we have new classes for non arrow layers that dont include fields like `table` or should we be extending existing classes and setting table to `null`/`None`

I'm not 100% sure yet, but at first glance I'd think to have another class `BaseArrowLayer` that's a subclass of `BaseLayer` and adds the `table` and `_rows_per_chunk` traits. Because everything other than those two is defined on deck's [base Layer class](https://deck.gl/docs/api-reference/core/layer), and should be common for all deck layers

PR #292, opened 2023-12-05
Implement `TripsLayer` for animating moving objects and connect to MovingPandas

kylebarron:
a minimal example
Screen Recording 2023-12-05 at 5 15 35 PM

### Change list

- Add new `TripsLayer` under the `experimental` module
- Add dev dependency on `movingpandas`
- Add `from_movingpandas` class method to construct a `TripsLayer` from a movingpandas `TrajectoryCollection`

### Todo

- [x] Implement conversion from `movingpandas.TrajectoryCollection` to GeoArrow.
- [x] More input validation in `TimestampAccessor`. Validate timestamps have the same offsetting as the geometry in the main data frame. (Done in https://github.com/developmentseed/lonboard/pull/292/commits/37a64c63a4c24ba4a194d7ced12da030fc7d726d)
- [x] Re-implement this example: https://movingpandas.github.io/movingpandas-website/2-analysis-examples/ship-data.html
- [x] Store a `time_offset` integer on the `TripsLayer` that represents the minimum value of the trip data. Note that you'd need to recompute this when a new `get_timestamps` is assigned onto the layer. (Done in https://github.com/developmentseed/lonboard/pull/292/commits/6addb2e04a4d1f886f6699f47c2384760f3dffc6)
- [x] Implement custom serialization for the timestamp accessor. Subtract off the time offset when serializing the data, and cast to float32. (Done in https://github.com/developmentseed/lonboard/pull/292/commits/b3468109af07d28aa67cd79b5490350e0dc45d52)
- [x] Add timezone parameter? (we infer the timezone from the input data)

### Open questions

- How to handle offsetted timestamps? deck.gl stores timestamps as float32, which means there isn't enough integer precision to store milliseconds or nanoseconds since epoch. (Timestamp precision handling done in https://github.com/developmentseed/lonboard/pull/292/commits/ca9dcd1939ccf0d0cd73efcbd9d2718aa38656f8)
- Where to handle animation? It looks like syncing animation via an `ipywidgets.Play` widget (connected via jslink) is probably good enough for now, even if it appears to have a decent amount of overhead. The alternative would be to have a manual animation component on the JS side that maintains its own time state.

### Example repro

I got data from [Access AIS](https://marinecadastre.gov/accessais/), with a custom bounding box and time range, though it would probably be straightforward to use [other data files](https://marinecadastre.gov/ais/) as well. 

```py
import pyarrow as pa
import pandas as pd
import movingpandas as mpd
from lonboard import Map
from lonboard.experimental import TripsLayer
import ipywidgets

path = '/Users/kyle/Downloads/AIS_170180417406763049_2306-1701804175229.csv'
df = pd.read_csv(path)
traj_collection = mpd.TrajectoryCollection(df, 'MMSI', t='BaseDateTime', x='LON', y='LAT')

layer.width_min_pixels = 5
layer.trail_length = 100000

play = ipywidgets.Play(
    value=0,
    min=0,
    max=86399000,
    step=50_000,
    interval=50,
    repeat=True
)
play
ipywidgets.jsdlink(
    (play, 'value'),
    (layer, 'current_time'),
)
```

cc @anitagraser, you may be interested in this, and/or have ideas for how to better integrate with movingpandas


anitagraser:
Thanks for tagging me. This development looks really exciting. Let me know if you have any movingpandas questions.

kdpenner:
Howdy, we talked at SciPi, posting to track this PR’s progress 🙂

lonboard/experimental/_layer.py
```diff
@@ -329,3 +357,216 @@ class TextLayer(BaseArrowLayer):
...
+
+        coords = np.zeros((num_coords, 2), dtype=np.float64)
+        timestamps = np.zeros(num_coords, dtype=np.int64)
+
+        for i, traj in enumerate(traj_collection.trajectories):
+            start_offset = offsets[i]
+            end_offset = offsets[i + 1]
+
+            # millisecond-based timestamps
+            int64_ms_timestamps = traj.df.index.to_series().astype(np.int64) // (
```
kdpenner, line 540: on the project I'm working on the timestamps in the parquet files (one every 3s) are stored at ms resolution so the next line creates duplicates. perhaps:
```suggestion
            int64_ms_timestamps = traj.df.index.as_unit("ns").to_series().astype(np.int64) // (
```
or a warning in the docs?

kylebarron: This PR no longer casts for the generic MovingPandas to GeoArrow conversion, and instead keeps whatever the existing time unit MovingPandas had. But we'll still have to do some conversion for Lonboard because the GPU uses float32 so we don't have enough precision to render nanosecond integers

lonboard/_geoarrow/movingpandas_interop.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def infer_time_unit(dtype: np.dtype) -> Tuple[Literal["s", "ms", "us", "ns"], DataType]:
+    """Infer an arrow time unit from the numpy data type
+
+    Raises:
+        ValueError: If not a known numpy datetime dtype
+    """
+
+    if dtype.name == "datetime64[s]":
```
jorisvandenbossche, line 164: Given that you are requiring pyarrow for `movingpandas_to_geoarrow` anyway, you could also use `pa.from_numpy_dtype(dtype)` and use the c interface to get the data type in arro3

Now, unfortunately that doesn't seem to handle pandas' custom tz-aware dtype .. (I was initially thinking this could be a solution to support those directly as well, but that's not the case; we should add some conversion function to pyarrow that supports any pandas dtype)

lonboard/_geoarrow/movingpandas_interop.py
```diff
@@ -0,0 +1,199 @@
...
+    time_unit, time_arrow_dtype = infer_time_unit(datetime_dtype)
+
+    # TODO: switch this to just using `time_arrow_dtype.bit_width` once
+    # https://github.com/kylebarron/arro3/pull/190 is released
+    if time_unit in {"s", "ms"}:
+        timestamps = np.zeros(num_coords, dtype=np.int32)
+    elif time_unit in {"us", "ns"}:
+        timestamps = np.zeros(num_coords, dtype=np.int64)
+    else:
+        raise ValueError(f"Unexpected time unit: {time_unit}.")
```
jorisvandenbossche, line 92: A timestamp type is always int64, regardless of the resolution? (that's for the time type where it depends)

lonboard/_geoarrow/movingpandas_interop.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # Loop second time to fill timestamps and coords
+    for i, traj in enumerate(traj_collection.trajectories):
+        start_offset = offsets[i]
+        end_offset = offsets[i + 1]
+
+        timestamps[start_offset:end_offset] = traj.df.index
+        coords[start_offset:end_offset, 0] = shapely.get_x(traj.df.geometry)
+        coords[start_offset:end_offset, 1] = shapely.get_y(traj.df.geometry)
+        coords[start_offset:end_offset, 2] = shapely.get_z(traj.df.geometry)
```
jorisvandenbossche, line 105:
```suggestion
        coords[start_offset:end_offset] = shapely.get_coordinates(traj.df.geometry, include_z=True)
```

This might work as well? (and be more efficient)

lonboard/_geoarrow/movingpandas_interop.py
```diff
@@ -0,0 +1,199 @@
...
+
+def apply_offsets_to_table(table: Table, offsets: Array) -> Table:
+    batch = table.combine_chunks().to_batches()[0]
+
+    new_fields = []
+    new_arrays = []
+
+    for field_idx in range(batch.num_columns):
+        field = batch.schema.field(field_idx)
+        new_field = field.with_type(DataType.list(field))
```
jorisvandenbossche, line 191: This relies on arro3's `list(..)` to convert the pyarrow field through the arrow pycapsule interface, and then pyarrow's `with_type()` convert again in the other direction? 
Nice ;)

jorisvandenbossche: Ah, no, I see that you converted the pyarrow Table to am arro3 one (`attr_table = Table.from_arrow(attr_table)`) before calling this function ;)

lonboard/_geoarrow/movingpandas_interop.py
kylebarron: Ah, thanks, indeed I mixed it up with the time type

lonboard/_geoarrow/movingpandas_interop.py
kylebarron: Here `batch` is actually an `arro3.RecordBatch`, the conversion back to arro3 happened here:
https://github.com/developmentseed/lonboard/blob/1ced62191ade5a110d2eef2df4524c5a5a49d740/lonboard/_geoarrow/movingpandas_interop.py#L116

But yes, these lines should work with pyarrow input as well!

lonboard/_geoarrow/movingpandas_interop.py
jorisvandenbossche: It's a bit of a mess, but I think the best option for now is something like this (using pyarrow for the moment):

```
def datetime64_dtype_to_arrow(dtype):
    if isinstance(dtype, pd.DatetimeTZDtype):
        return pa.timestamp(dtype.unit, dtype.tz)
    else:
        return pa.from_numpy_dtype(dtype)
```

if you don't want to rely on pyarrow and create the timestamp type manually, you could still use the comparison with `dtype.name` as you do now

lonboard/_geoarrow/movingpandas_interop.py
kylebarron: > Ah, no, I see that you converted the pyarrow Table to am arro3 one (`attr_table = Table.from_arrow(attr_table)`) before calling this function ;)

Right. I could've left it as a pyarrow table; the main difference is mostly for type hinting. It's nice to get IDE completions and type checking and pyarrow doesn't have that yet.

lonboard/_geoarrow/movingpandas_interop.py
kylebarron: Implemented in https://github.com/developmentseed/lonboard/pull/292/commits/dd56ef3c3ee2d6df02771ac473cb60ca18dece05

lonboard/_geoarrow/movingpandas_interop.py
jorisvandenbossche: > Interestingly... `shapely.get_coordinates` is 3x faster than `get_x` in isolation!

I was going to say that this is expected, because `get_coordinates` can do cheaper memcpy of the full coordinate sequence buffer. But, I would actually only expect to see a significant difference there for geometries with longer CoordSeqs, so not for points ..

Now, if I test this with a random array, I do see that an individual `get_x` is faster:

```
In [10]: arr = shapely.points(np.random.randn(10000, 2))

In [11]: %timeit shapely.get_x(arr)
256 µs ± 25.3 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [13]: %timeit shapely.get_coordinates(arr)
966 µs ± 117 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

So in practice I think what you are seeing above is mostly a slightly different overhead between both options for small arrays. In the end, it's probably only a code simplication and so not much of a performance improvement

jorisvandenbossche: > TIL that there's a considerable difference between applying shapely operations on a `GeometryArray` vs a `GeoSeries`:

Yeah, that is because in the GeoSeries case it reconstructs a (Geo)Series from the result (at least for ufuncs like `get_x`, not for `get_coordinates`), which has some overhead. And in this case you don't need a series, but just an array and so can avoid that overhead by calling the shapely function on the array

kylebarron:
I added `min` and `sub` operations to arro3 https://github.com/kylebarron/arro3/pull/193 and https://github.com/kylebarron/arro3/pull/194.

So for storing the timestamp offset (minimum timestamp in the data), we'll use the `min` kernel. And then in the serialization process we'll use `sub` to subtract that min timestamp off, and lastly cast to `float32` before sending to the browser.

By storing the min timestamp as an arrow scalar, we should be able to maintain the original precision and metadata of the input time data. This should make it easier to create a time slider and animation control that use real string-formatted times instead of integers from epoch.

lonboard/experimental/_layer.py
kylebarron: @kdpenner you might be interested in looking at [`ca9dcd1` (#292)](https://github.com/developmentseed/lonboard/pull/292/commits/ca9dcd1939ccf0d0cd73efcbd9d2718aa38656f8). Because deck.gl stores timestamps as float32 values, we're forced to ensure that the input data can be represented as integers within the float32 integer range. (Technically we don't need _integers_ within float32 values, but we're losing precision _somewhere_ regardless).

So if you pass in data with nanosecond precision and it has a larger range than can fit in float32 integers (33554432) then a warning will be printed and the input data will automatically be reduced in precision, as far as to seconds.

kylebarron:
I think this is almost ready to go!

Tasks:

- [ ] Publish new version of arro3 with necessary updates from https://github.com/kylebarron/arro3/pull/199
- [ ] Create `play_widget()` as a method on the `TripsLayer` class. Potentially the `step` can even be a scalar representing a duration? And then I map that step to the integer step?
- [ ] Have a helper function to go from the integer shown on the widget back to a timestamp.
- [ ] Implement example from movingpandas

kylebarron:
I'm considering taking this out of the `experimental` module because the core behavior is pretty stable. It's just the interop with external libraries (other than movingpandas) that isn't really stable.

kylebarron:
I'm not sure on the best API regarding `current_time`. The last commit created a `current_time_as_datetime` function to convert from integer back to a `datetime` object. But really `current_time`, despite being a public deck.gl API, is an internal construct for the Lonboard `TripsLayer`. So we should probably make `current_time` private as `_current_time`. And since we manage the animation details in `animate()`, that should be fine. Then should there be a `current_time` method (getter?) that does the conversion from `_current_time` as int to a `datetime` object?

We should also check how `datetime` works with time zones.

kylebarron:
Got an air traffic control example working too:

https://github.com/user-attachments/assets/6ab3706e-525e-4b5a-9904-e126d043cc2c

https://github.com/user-attachments/assets/9d7643a1-63bc-4236-bcd0-8412ff2f25d9

On Monday we can just clean up the examples a bit, add a display for the current time of the animation, and then publish the new version!

kdpenner:
I'm able to view ~4 million points on my laptop with reduced performance. 800K look great. if you are looking for more examples, here are open data affiliated with our project: https://osf.io/dg6t3/

the `trajectories_*` subfolders have (unfortunately only) 1 timestamp every 5 minutes.

for the wish list: will be great when the `from_geopandas` inherited class method works with native datetime or pandas timestamp objects 🙂 

cc @hengoren

lonboard/experimental/_layer.py
```diff
@@ -329,3 +353,333 @@ class TextLayer(BaseArrowLayer):
     default [0, 0]
     """
     #  ?: Accessor<DataT, [number, number]>;
+
+
+class TripsLayer(BaseArrowLayer):
+    """The `TripsLayer` renders animated paths that represent vehicle trips.
```
kdpenner, line 359: not just vehicles!

kylebarron: Generally I just copy that docstring from deck.gl (https://deck.gl/docs/api-reference/geo-layers/trips-layer), but in this case maybe we should choose a different description

kylebarron:
@kdpenner if you'd like to create an example notebook in your own repo, we can link to it from our docs

kylebarron:
> for the wish list: will be great when the `from_geopandas` inherited class method works with native datetime or pandas timestamp objects 🙂

It is technically possible, though difficult, to use `from_geopandas` or `from_duckdb` with the `TripsLayer`, but you have to pass in the `get_timestamps` parameter separately, and ensure the list sizes match the LineString geometries.

kdpenner:
FWIW `layer.get_timestamps.to_numpy()` raises a `NotImplementedError`:

> NotImplementedError: Unsupported type in to_numpy List(Field { name: "", data_type: Timestamp(Second, None), nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} })

`to_pylist()` works

PR #315, opened 2024-01-11
Fix CI

kylebarron:
@emmalu It's best if you use pull requests instead of committing directly to the `main` branch. There's some lint checking that runs even on Markdown files, so your commit broke the CI https://github.com/developmentseed/lonboard/actions/runs/7490075262/job/20388095204


emmalu:
Aw, bummer. And of course I hesitated, but wanted to get the template in as soon as I could. Noted though! Thanks for the note. Will not happen again! 👍🏽

Issue #322, opened 2024-01-17
PointCloudLayer


kylebarron:
@naomatheus was asking about using the PointCloudLayer in lonboard. 

- Expose multiple views to Python, so that advanced users can choose alternate views
- Add PointCloudLayer to deck.gl-layers and then to lonboard. This is similar in most respects to the ScatterplotLayer, but it also has a [`getNormal` accessor](https://deck.gl/docs/api-reference/layers/point-cloud-layer) for point clouds specifically. (I have to do more research to understand exactly how this would be used)
- Add example connecting to pdal

Note that the layer itself and the views are two distinct streams of work.
```[tasklist]
### Tasks
- [x] add PointCloudLayer to deck.gl-layers
- [x] Add validator for NormalAccessor
- [x] Create new python PointCloudLayer
- [ ] Create example using PDAL
```

kylebarron:
Relevant comment for adding a new layer to lonboard: https://github.com/developmentseed/lonboard/issues/302#issuecomment-1874423165

For PointCloudLayer, for `getNormal` we want to have a new type of "accessor" where the type is an Arrow `FixedSizeList` with size `3` and data type `float32`. Then we can pass that binary buffer into the [`getNormal`](https://deck.gl/docs/api-reference/layers/point-cloud-layer#getnormal) param on the layer.

So in deck.gl-layers, we can create a new type called `NormalAccessor` that's defined as `arrow.Vector<arrow.FixedSizeList<arrow.Float32>>`.
https://github.com/geoarrow/deck.gl-layers/blob/b8cd17ad1b956420199b69faf138a44ae4d19e32/src/types.ts#L12-L14

Then create a new file, matching the types of the `GeoArrowPointCloudLayer` to the upstream layer props. E.g. with the scatterplot layer you can see [our types](https://github.com/geoarrow/deck.gl-layers/blob/b8cd17ad1b956420199b69faf138a44ae4d19e32/src/scatterplot-layer.ts#L37-L72) and how they relate to the [upstream parameters](https://deck.gl/docs/api-reference/layers/scatterplot-layer).

First create a PR there that we can discuss, and then after that we can go through how to add a new validator for this `NormalAccessor` data type, which allows us to create a new Python `PointCloudLayer`.

kylebarron:
[...]

PR #328, opened 2024-01-26
DRAFT: add validation in python for NormalAccessor

naomatheus:
Draft PR to add NormalAccessor validation to Python experimental traits in `lonboard`.

I had some unrelated git history issues as I renamed the local repository to help stay organized, hence the branch name.

### DIFF
--- a/lonboard/experimental/traits.py
+++ b/lonboard/experimental/traits.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from typing import Any, List, Tuple, Union
 
 import matplotlib as mpl
@@ -7,10 +8,7 @@ import numpy as np
 import pyarrow as pa
 from traitlets.traitlets import TraitType
 
-from lonboard._serialization import (
-    COLOR_SERIALIZATION,
-    NORMAL_SERIALIZATION
-)
+from lonboard._serialization import COLOR_SERIALIZATION, NORMAL_SERIALIZATION
 from lonboard.traits import FixedErrorTraitType
 
 
@@ -106,14 +104,15 @@ class PointAccessor(FixedErrorTraitType):
 
 class NormalAccessor(Union[TraitType, FixedErrorTraitType]):
     """
-        A representation of a deck.gl normal accessor
-        
-        Acceptable inputs:
-        - A numpy ndarray with two dimensions: The size of the second dimension must be 3 i.e. `(N,3)`
-        - a pyarrow `FixedSizeListArray` or `ChunkedArray` containing `FixedSizeListArray`s 
-        where the size of the inner fixed size list 3.
+    A representation of a deck.gl normal accessor
+
+    Acceptable inputs:
+    - A numpy ndarray with two dimensions: The size of the second dimension must be 3 i.e. `(N,3)`
+    - a pyarrow `FixedSizeListArray` or `ChunkedArray` containing `FixedSizeListArray`s
+    where the size of the inner fixed size list 3.
     """
-    default_value = [0,0,1]
+
+    default_value = [0, 0, 1]
     info_text = (
         "List representing normal of each object, in [nx, ny, nz]. or numpy ndarray or "
         "pyarrow FixedSizeList representing the normal of each object, in [nx, ny, nz]"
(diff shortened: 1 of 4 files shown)


naomatheus:
This could potentially be quite off base. But just took a stab at it for a draft.

This PR was created in response to reference here: https://github.com/developmentseed/lonboard/issues/322#issuecomment-1896441765

| quote test
|

kylebarron:
We should add a mention of the pre-commit setup in DEVELOP.md, but you need to run
```
pip install -U pre-commit
pre-commit run --all-files
```
Your PR is erroring on formatting.

lonboard/experimental/traits.py
```diff
@@ -101,3 +102,71 @@ def validate(
...
+        
+        if isinstance(value, np.ndarray):
+            # values must be contiguous (the same length for all values)
+            if value.ndim != 2 or value.shape[1] != 3:
+                self.error(obj, value, info="Normal array must be 2D with shape (N,3)")
+                
+            if not np.any(value == 1):
+                self.error(obj, value, info="One of the normal vector elements must be 1.")
+                
+            return pa.FixedSizeListArray.from_arrays(value.flatten("C"), list_size)
```
kylebarron, line 152: We need to ensure that numpy arrays are floating point data type. You can check with `np.issubdtype(value.dtype, np.floating)` and then we want to cast it to float32 if it's not already.
```py
if not np.issubdtype(value.dtype, np.float32):
    value.astype(np.float32)
```

lonboard/experimental/traits.py
```diff
@@ -101,3 +102,71 @@ def validate(
...
+            if value.type.list_size != 3:
+                self.error(
+                    obj,
+                    value,
+                    info=(
+                        "Normal pyarrow array must have a FixedSizeList inner size of 3."
+                    ),
+                )
+
+            return pa.FixedSizeListArray.from_arrays(value.flatten("C"), list_size)
```
kylebarron, line 169: Same thing with pyarrow arrays, we need to validate that it's of a floating point data type, and then cast it to float32. It may be easier in the numpy section to overwrite the `value` variable with a pyarrow array, which will then activate this `if` block, and you only have to cast data types once.

lonboard/experimental/traits.py
```diff
@@ -101,3 +102,71 @@ def validate(
...
+
+    def __init__(
+        self: TraitType,
+        *args,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.tag(sync=True, **NORMAL_SERIALIZATION)
+        
+    def validate(
```
kylebarron, line 130: I'd recommend writing some tests first, which will make it easier to sort out the logic in this function. You can find some related tests [here](https://github.com/developmentseed/lonboard/blob/16d4c3ce41fbd540585b4ef3e5395cee4038e8ee/tests/test_traits.py).

The purpose of this validation function is to take whatever user input and both validate it _and_ conform to a standard expectation of how we want to store it internally.

To align with deck.gl and other traits, we should allow a list with three scalar values, or a numpy array, or a pyarrow FixedSizeListArray. 

But here the first thing you check is `value.shape`. Which won't exist for a list type. So it'll crash

```
In [1]: l = [1, 23, 50]

In [2]: l.shape
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 l.shape

AttributeError: 'list' object has no attribute 'shape'
```

Hence why adding tests first might help

kylebarron (changes requested):

lonboard/experimental/traits.py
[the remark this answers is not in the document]
kylebarron: the pyarrow `FixedSizeListArray` takes a flat buffer as input, which it then "reshapes" into a 2d array with width `list_size`

lonboard/experimental/traits.py
```diff
@@ -95,9 +94,104 @@ def validate(
...
+            return pa.FixedSizeListArray.from_arrays(value, fixed_list_size)
+
+        if isinstance(value, np.ndarray):
+            if value.ndim != 2 or value.shape[1] != 3:
+                self.error(obj, value, info="Normal array must be 2D with shape (N,3)")
+
+            if not np.any(value == 1):
+                self.error(
+                    obj, value, info="One of the normal vector elements must be 1."
+                )
```
kylebarron, line 156: This is a great idea for a validation step, but I don't _think_ it's correct. There's no good documentation in deck.gl of the exact format expected. Most likely, these normal vectors are expected to be "normalized" which means that their _length_ is expected to be 1. I.e. x^2 + y^2 + z^2 = 1. This means that almost never will one of the elements themselves be 1.

In any case I'd tend to think this is something we shouldn't validate, because deck.gl might be lenient and allow non-normalized vectors.

lonboard/experimental/traits.py
```diff
@@ -95,9 +94,104 @@ def validate(
...
+                    value,
+                    info=(
+                        """Normal pyarrow array must have a
+                        FixedSizeList inner size of 3."""
+                    ),
+                )
+
+            if not pa.types.is_floating(value.type.value_type):
+                try:
+                    value = value.cast(pa.list_(pa.float32()))
```
kylebarron, line 185: Ditto on the `pa.list_` constructor

lonboard/experimental/traits.py
```diff
@@ -87,7 +85,7 @@ def validate(
         if isinstance(value, str):
             try:
                 c = mpl.colors.to_rgba(value)  # type: ignore
-            except ValueError:
+            except Exception:
```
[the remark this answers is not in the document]
naomatheus: Wrote this while blocked on testing so this hasn't been executed

lonboard/experimental/traits.py
```diff
@@ -126,47 +126,72 @@ def __init__(
...
+        (the same length for all values) and of floating point type
+        """
+
+        fixed_list_size = 3
+
+        if isinstance(value, List):
+            if len(value) != 3:
+                self.error("Normal array must have length 3, (x,y,z)")
+
+            if not all(isinstance(item, float) for item in value):
```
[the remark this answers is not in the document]
naomatheus: I thought [0,0,1] would be invalid and only [0,0,1.0] would be valid

naomatheus:
Hey @kylebarron I will start a new branch that's better synced with `main`

PR #330, opened 2024-01-26
WIP: PolygonLayer

kylebarron:
It turns out that there are still some issues on the JS side of the PolygonLayer. I think I had only tested it with Polygon, not MultiPolygon, input, which is why I didn't catch these two. 

More info in https://github.com/geoarrow/deck.gl-layers/issues/102


kylebarron:
Example of default colors in `viz()`:

image

PR #341, opened 2024-01-31
Bump parquet and use arrow-js-ffi

kylebarron:


kylebarron:
Note that we only need a param for the record batch size. In Lonboard we already send separate Parquet files that each contain a single logical Arrow batch. So we really just want to set the  record batch size to a very large number.

kylebarron:
I published parquet-wasm 0.6.0 and it would be good to test this again. We should bump parquet-wasm regardless of whether we enable arrow-js-ffi.

kylebarron:
Superseded by https://github.com/developmentseed/lonboard/pull/477

PR #351, opened 2024-02-08
adds an example using PolygonLayer

naomatheus:
Added a short example notebook using the PolygonLayer together to create a map of municipal boundaries. 

~TODO: I'll need to add images and/or gifs to make the example more visually appealing.~


naomatheus:
I'd love to add an image or a gif to show the result at the end of the notebook. Any preferred way to do that?
I suppose I would just add an image into the examples directory. Or perhaps a separate directory for the images.
And then link from the notebook with `! [alt_text] path_to_image`

Something like that

kylebarron:
You can put an image into the `assets/` directory. Try to make it relatively small file size. You can do something like `brew install imagemagick; convert input.png output.jpg --quality 75`. (I forget the exact syntax; that might be wrong)

kylebarron:
I'd forgotten that I hadn't given this a full review (feel free to ping me if I haven't reviewed in a day or two), though it would be great to have a screenshot for this as well.

Since it's docs only, this can be merged after the release

naomatheus:
okay! Moving this out of draft

naomatheus:
I'm curious how the gif holds up as well

examples/global-boundaries.ipynb
```diff
@@ -0,0 +1,648 @@
...
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "# Use pyogrio engine for an optimization of the time it takes to read this file\n",
+    "# We will also enable use_arrow, for speed\n",
+    "\n",
+    "# this bounding box will visualize the Western Hemisphere - North, Central, and South America.\n",
+    "nh_bbox = (-175.910339,-59.265881,-27.199402,74.449358)\n",
```
kylebarron, line 135: I'm curious how many coordinates are in the dataset if we remove this bounding box. Is it too much data?

examples/global-boundaries.ipynb
```diff
@@ -0,0 +1,648 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 37,
+   "id": "5bee50cf-1b99-4fa0-b484-d4c2d0ca6bc8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Step 1: Calculate 'boundary' for each geometry\n",
+    "# Instead of adding it to the original gdf, directly create a new GeoSeries\n",
+    "boundaries = gdf['geometry'].apply(lambda geom: geom.boundary)\n",
```
kylebarron, line 455: [`shapely.boundary`](https://shapely.readthedocs.io/en/stable/reference/shapely.boundary.html) will probably be much faster calling `.apply`, and is a better option to suggest to users. This is because `geom.boundary` operates on _one geometry at a time_, while `shapely.boundary()`  operates on the entire array at once, which is a lot faster.

examples/global-boundaries.ipynb
```diff
@@ -0,0 +1,648 @@
...
+    "mississippi = gdf[gdf['shapeName'] == 'Mississippi']\n",
+    "mississippi"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a899d398-ecb2-4557-b803-71e0c0a42470",
+   "metadata": {},
+   "source": [
+    "For MultiPolygons, we can get the `unary_union` of the polygon to obtain a `boundary` of the `MultiPolygon`"
```
kylebarron, line 372: Here we introduce the idea of `unary_union`, but then it's never used below when constructing the boundary. It might be easier to follow if we removed this `unary_union` example. What do you think?

examples/global-boundaries.ipynb
```diff
@@ -0,0 +1,648 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a3f4b54a-44b3-4033-b459-37f84df2f49a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# set up our fill and line colors\n",
+    "fill_rgb = [204, 251, 254] # light blue\n",
+    "border_rgb = [37, 36, 34] # dark border color"
```
kylebarron, line 561: Are there any attributes in the data that might make sense for rendering a different color per polygon?

kylebarron (commented):
This looks pretty good! Just a few comments

examples/global-boundaries.ipynb
```diff
@@ -0,0 +1,648 @@
...
+   "cell_type": "markdown",
+   "id": "a03ebb10-7317-4b90-ab5e-9da56a37a9da",
+   "metadata": {},
+   "source": [
+    "# Municipal Boundaries in the Americas\n",
+    "\n",
+    "This example will use data from GeoBoundaries.org's data set of [Comprehensive Global Administrative Zones (CGAZ)] (https://www.geoboundaries.org/globalDownloads.html).\n",
+    "\n",
+    "> CGAZ is a set of global composites for administrative boundaries. There are two important distinctions between our global product and individual country downloads. First, extensive simplification is performed to ensure that file sizes are small enough to be used in most traditional desktop software. Second, disputed areas are removed and replaced with polygons following US Department of State definitions.\n",
+    "> Attribution: www.geoboundaries.org"
```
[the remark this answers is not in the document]
naomatheus: added a specific note to the top of this notebook

examples/global-boundaries.ipynb
naomatheus: It looks like there are ~3224 rows. But some of those rows contain multipolygons, so more than one boundary in one. Others are single polygons with one polygon per admin boundary.

naomatheus: Yea actually upon running this a second time there's not a significant difference in the time it takes to load the data with vs without the northern hemisphere bounding box. So I have removed that.

examples/global-boundaries.ipynb
naomatheus: Thanks. I will swap this out as well and use `shapely.boundary` instead. 
As I was first making the example, I was getting AttributeError when running `geom.boundary` on the multipolygon geometries. So my thought process was find a way to access the `boundary` attribute on both multiply and singlepoly geom types.

examples/global-boundaries.ipynb
naomatheus: Sounds good. Removed via https://github.com/developmentseed/lonboard/pull/351#discussion_r1503471224

examples/global-boundaries.ipynb
naomatheus: The only thing I can think of is... we could possibly render a different color for each `ShapeGroup`.
The shapeGroup attribute is roughly a "country code".
The first features in the FeatureCollection are in `shapeGroup` AFG for Afghanistan.

I think it is pretty cool so quickly render a map of the entire world (like whoa). But I could add some additional logic to show off features like allowing a specific render color based on the "country code"

What do you think?

naomatheus:
Hmm... One comment I have here is that I noticed that rendering the entire global map seems to affect the rendering of the `fill_color` and the map's pickability. When I render the entire map, it is no longer pickable for some reason.

examples/global-boundaries.ipynb
emmalu: That `ShapeGroup` property seems like the most viable choice to me (there isn't a ton to choose from, is there?) But it would bring a little clarity to the geometries (countries), especially those that are close together in Europe/Middle East.

emmalu (approved):
Nice notebook example, Matt. Especially if you can distinguish shape colors by country code. 👍🏽

kylebarron:
There are some weird artifacts here:
image

It's hard to know whether that's an issue in the original data or an issue with `shapely.boundary` or something else. In any case I do also see that the rendering and picking is somewhat slow on my computer. That said, it is 8 million coordinates, so that is quite a lot to visualize!

How about we bring back in the spatial filter over the americas for now? What do you think?

kylebarron:
boundaries

Updated this to use the new PolygonLayer

PR #376, opened 2024-02-20
NormalAccessor

naomatheus:
Adds NormalAccessor and draft tests
Respond to comments on #328 and move recent work here. Branch #328 will be deleted


lonboard/traits.py
```diff
@@ -705,3 +706,99 @@ def validate(self, obj, value) -> Union[float, pa.ChunkedArray, pa.DoubleArray]:
 
         self.error(obj, value)
         assert False
+
+
+class NormalAccessor(FixedErrorTraitType):
+    """
+    A representation of a deck.gl normal accessor
+
+    Acceptable inputs:
```
kylebarron, line 715: We should also document that you can pass a list/tuple of length 3.

lonboard/traits.py
```diff
@@ -705,3 +706,99 @@ def validate(self, obj, value) -> Union[float, pa.ChunkedArray, pa.DoubleArray]:
...
+    - A numpy ndarray with two dimensions: The size of the
+    second dimension must be 3 i.e. `(N,3)`
+    - a pyarrow `FixedSizeListArray` or
+    `ChunkedArray` containing `FixedSizeListArray`s
+    where the size of the inner fixed size list 3.
+    """
+
+    default_value = [0, 0, 1]
+    info_text = (
+        "List representing normal of each object, in [nx, ny, nz]. or numpy ndarray or "
```
kylebarron, line 725:
```suggestion
        "List representing normal of all objects in [nx, ny, nz] or numpy ndarray or "
```

lonboard/traits.py
```diff
@@ -705,3 +706,99 @@ def validate(self, obj, value) -> Union[float, pa.ChunkedArray, pa.DoubleArray]:
...
+                self.error(obj, value, info="Normal array must be 2D with shape (N,3)")
+
+            if not np.any(value == 1):
+                self.error(
+                    obj, value, info="One of the normal vector elements must be 1."
+                )
+
+            if not np.issubdtype(value.dtype, np.float32):
+                warnings.warn(
+                    """Warning: Numpy array should be floating point type.
```
kylebarron, line 767:
```suggestion
                    """Warning: Numpy array should be float32 type.
```

lonboard/traits.py
```diff
@@ -705,3 +706,99 @@ def validate(self, obj, value) -> Union[float, pa.ChunkedArray, pa.DoubleArray]:
...
+                    Converting to float32 point pyarrow array"""
+                )
+                value = value.astype(np.float(pa.list_(pa.float32(), 3)))
```
kylebarron, line 770: This will fail; you want `value.astype(np.float32)`. You can't pass a pyarrow type into a numpy function like this.

lonboard/traits.py
```diff
@@ -705,3 +706,99 @@ def validate(self, obj, value) -> Union[float, pa.ChunkedArray, pa.DoubleArray]:
...
+
+            return pa.FixedSizeListArray.from_arrays(value, fixed_list_size)
+
+        if isinstance(value, (pa.ChunkedArray, pa.Array)):
+            if not pa.types.is_fixed_size_list(value.type):
+                self.error(
+                    obj, value, info="Point pyarrow array must be a FixedSizeList."
```
kylebarron, line 777:
```suggestion
                    obj, value, info="normal pyarrow array must be a FixedSizeList."
```

tests/test_traits.py
```diff
@@ -343,3 +348,70 @@ def test_filter_value_validation_filter_size_3():
...
+
+    value = NormalAccessor()
+
+
+def test_normal_accessor_validation_list_length():
+    # tuple or list must have 3 elements
+    with pytest.raises(TraitError):
+        NormalAccessorWidget()
+
+    with pytest.raises(TraitError, match="expected 3 values if passed a tuple or list"):
```
kylebarron, line 366: The `match` here is tested against the error string, so if you're copying tests from above you might need to adjust the expected error string, or use the same error message in your code.

tests/test_traits.py
```diff
@@ -343,3 +348,70 @@ def test_filter_value_validation_filter_size_3():
...
+        NormalAccessorWidget(value=(1, 2))
+
+    with pytest.raises(TraitError, match="expected 3 values if passed a tuple or list"):
+        NormalAccessorWidget(value=(1, 2, 3, 4))
+    NormalAccessorWidget(value=(1, 2, 3))
+    NormalAccessorWidget(value=(1, 2, 3))
```
kylebarron, line 372: These two lines are the same; you can delete one of them.

tests/test_traits.py
```diff
@@ -343,3 +348,70 @@ def test_filter_value_validation_filter_size_3():
...
+    with pytest.raises(TraitError, match="expected scalar values"):
+        NormalAccessorWidget(value=["1.1", 2.2, 3.3])
+    NormalAccessorWidget(value=[1.1, 2.2, 3.3])
+
+
+def test_normal_accessor_validation_dim_shape_np_arr():
+    # expected value is a numpy ndarray with two dimensions
+    # And  size of the second dimension must be 3 i.e. `(N,3)`
+    with pytest.raises(TraitError, match="expected shape (N,3)"):
+        NormalAccessorWidget(value=np.array([1, 2], 3))
```
kylebarron, line 386: All of these numpy calls will fail.

To construct a 2d numpy array you can do either
```py
np.array([[1, 2], [3, 4]])
```
which gives a 2x2 numpy array or 
```
np.array([1, 2, 3, 4, 5, 6]).reshape(-1, 2)
```
which first creates a 1-dimensional array and then rearranges into a 2-d array, in this case of shape `(3, 2)`.

naomatheus: Right. I'll bring this as a general testing question because I thought that we were intentionally creating fail conditions and ensuring that the `NormalAcc` fails on these objects (like where the shape is just incorrect).

tests/test_traits.py
naomatheus: Yes I will update these to match

tests/test_traits.py
kylebarron: Yeah I mean that we want to test that passing a _valid_ numpy array into `NormalAccessorWidget` fails. If the numpy creation fails before passing into our code, we haven't tested our own code yet.

kylebarron:
Thanks! I cleaned this up and updated the tests

PR #379, opened 2024-02-23
Add CLI

kylebarron:
Depends on https://github.com/developmentseed/lonboard/pull/372

Needs docs; hasn't been tested yet


lonboard/_cli.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # If -o flag passed, write to file; otherwise write to stdout
+    if output:
+        map_.to_html(output)
+    else:
+        map_.to_html(sys.stdout)
+
+
+if __name__ == "__main__":
+    main()
```
vincentsarago, line 108: this is only need if you want to support something like `python -m lonboard._cli.main` but it's not mandator y

IMO, cli should be only handled at pyproject level with https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#creating-executable-scripts

lonboard/__main__.py
```diff
@@ -0,0 +1,3 @@
+from lonboard._cli import main
+
+main()
```
vincentsarago, line 3: not sure why we need this

pyproject.toml
```diff
@@ -19,7 +19,11 @@ palettable = "^3.3.3"
 # future to remove the matplotlib dependency.
 matplotlib = "^3.7"
 shapely = "^2"
+click = { version = "^8.1.7", optional = true }
+pyogrio = { version = "^0.7.2", optional = true }
```
vincentsarago, line 23: should we make pyogrio an optional dependency?

lonboard/_cli.py
```diff
@@ -0,0 +1,108 @@
+import json
+import sys
+from pathlib import Path
+from typing import Dict, List, Optional
+
+import click
+import pyarrow as pa
+import pyarrow.parquet as pq
+from pyogrio.raw import read_arrow
```
vincentsarago, line 9: because `pyogrio` is an optional dependency we should do something like 

```
try:
    import pyogrio
except ImportError:  # pragma: nocover
    pyogrio = None  # type: ignore
```

and at L93 raise an error with `pyogrio` is None

lonboard/_cli.py
```diff
@@ -0,0 +1,108 @@
...
+
+        tables.append(table)
+
+    map_ = viz(tables)
+
+    # If -o flag passed, write to file; otherwise write to stdout
+    if output:
+        map_.to_html(output)
+    else:
+        map_.to_html(sys.stdout)
```
vincentsarago, line 104: in one project I've done something like 


```python
import tempfile


if not output:
    with tempfile.NamedTemporaryFile(suffix=".html", delete=False) as f:
        output = f.name

map_.to_html(f.name)

click.launch(output)
```

tempfile will automatically create a temporary file in `/tmp` which will get deleted by the system at one point. 

if we go this way we could add a warning printing the path to the file.

lonboard/_cli.py
kylebarron: I figured that I'd make it required for the CLI and be ok with an `ImportError`, but I suppose that we could allow the CLI to work for GeoParquet files without pyogrio installed

lonboard/__main__.py
kylebarron: This enables `python -m lonboard file.geojson`

lonboard/_cli.py
kylebarron: Yeah we should just remove this. And yes, I do need to add the entry point into the `pyproject.toml`.

PR #399, opened 2024-03-01
Add type hinting for constructors and `from_geopandas`

kylebarron:
Better type hinting in vscode! This should hopefully be a massive QOL upgrade, at least in vscode. This doesn't show up in Jupyter though 🤷‍♂️ 

image
image


kylebarron:
Note to self, I wanted to do this with generics, so that I didn't have to add the two functions to each subclass, but that appears not to work: https://github.com/microsoft/pyright/issues/5574

PR #411, opened 2024-03-08
CLI: 'crs' in geoparquet metadata should be optional.

jwass:
GeoParquet spec allows crs to be optional in the metadata. If missing don't pass it through the GeoArrow metadata.

Tested with a geoparquet file with a missing crs. Displays it with this fix.


kylebarron:
Thanks!

Issue #405, opened 2024-03-04
Record and export animations


kylebarron:
A powerful feature would be to export image stills or video screencasts from a map. This would be really exciting for power users to make e.g. content for presentations.

There's a JS library called [hubble.gl](https://hubble.gl/) that creates animations from deck.gl maps. For example the following video is created from [this example](https://hubble.gl/examples/nyc-trips/).

[4bf79c3f-ba20-c98b-927c-0be9184f08d6.webm](https://github.com/developmentseed/lonboard/assets/15164633/410ee9f9-df3a-4423-85af-b9e40b174706)

In theory this might not be too hard to support. The [basic animation example](https://hubble.gl/docs#basic-animation) is pretty minimal, and most importantly the _user input_ is all json-serializable, so it would be relatively easy to expose to a user. 

After creating the video, in theory we could send the data back from the browser to Python, but it seems a lot easier to just "download" the video from the user's browser.

batpad:
This is how it's being done in `gl-director` - if one just wants the user to hit "Record", do what they want with the map, "Stop Recording" and then be able to download a video file: https://github.com/developmentseed/gl-director/blob/main/app/assets/scripts/components/home/video-pane.js#L62 - I think that same approach should work.

kylebarron:
Note that I would guess that uses a different mechanism than what hubble.gl is using. In particular, hubble.gl hooks directly into the map itself. This means that it can ensure all the data has been loaded and rendered for every keyframe. So the generated animation looks perfectly smooth.

I'm guessing that if you used gl-director, it records the map just the same as if you were manually dragging the map?

batpad:
> I'm guessing that if you used gl-director, it records the map just the same as if you were manually dragging the map?

Yea, gl-director hooks directly onto the canvas element so would just record whatever is happening there, and yes, this would likely replicate any rendering lag in the video.

Hubble.gl looks cool - if you think we could hook that up and give more granular control over frame-rate, etc. that does sound likely better.

kylebarron:
I think it's definitely possible to connect it to Lonboard. It might take a day or two to explore how the integration would work

PR #417, opened 2024-03-13
Select by bounding box

kylebarron:
Building on top of https://github.com/developmentseed/lonboard/pull/412 with help from @batpad

### Change list

- Draw bbox in deck instead of separate layer

### Notes

- Create mode for when picking bbox should be active
- Send data back to Python
- Make multiple draw calls, one per set of layer ids, then set on layer state
- Maybe delegate to sub models directly. Add a `pick` callback directly onto the models, and then iterate over each model, calling `pick` which picks and then assigns directly onto that model's state?
- ~In testing, we seemed to get only 35k rows back when we got 143k rows checking the same bbox in geopandas with the input gdf. This is possibly related to x,y,width, height passed in to `pickObjects`?~ Fixed with latest commit
- Note about crossing international dateline
- Ensure we're not creating more than 256 layers.


batpad:
@kylebarron I added a "selectionMode" and some ugly UI for a user to start selection mode and clear selection. As a bonus, when the user has a selection, it also displays the number of objects selected.

There's a few things I feel a bit unsure about the logic, and also how the code is structured. We can look tomorrow - but I think this does roughly what we'd want.

kylebarron:
It probably makes sense to keep the bounding box state on the widget model so that Python has access to the same bounding box. So e.g. connecting to Mosaic you could pass the bounding box for them to do the same query in duckdb

batpad:
There would still a few things to be done here to finish, here's what it looks like so far:

https://github.com/developmentseed/lonboard/assets/72280/c7e85f75-e9eb-4bdc-be74-21cc40059b91

@kylebarron - Although, I'm wondering if this can be restructured to not be so tightly coupled to the main rendering code in `index.tsx` - it would be almost nice to be able to build something like this as a separate widget, not inside the core `lonboard` code-base. I think that's probably doable - it might be nice to talk through next week.

cc @hanbyul-here

src/index.tsx
```diff
@@ -91,7 +98,7 @@ function App() {
   });
 
   const [mapId] = useState(uuidv4());
-
+  let mapRef = useRef<DeckGLRef>(null);
```
isaacbrodsky, line 101: nit: this and the useState above can all be const

src/index.tsx
```diff
@@ -144,8 +151,153 @@ function App() {
...
+        data,
+        filled: true,
+        getFillColor: [0, 0, 0, 30],
+        stroked: true,
+        getLineWidth: 2,
+        lineWidthUnits: "pixels",
+        getPolygon: (d) => d.polygon,
+      });
+    } else if (selectionStart) {
+      // Show the selection start point (note this does not show the proposed bounding box, but could be done)
```
isaacbrodsky, line 254:
```suggestion
      // Show the selection start point
```
Since a hover box is now shown

src/index.tsx
kylebarron: I think this is me writing too much Rust and always using `let`...

batpad:
@kylebarron for now, this just sets `map.selected_bounds` after the user selects the bbox on the python side.

As an approach, I think I prefer this than trying to send indexes for all the objects down from JS - can we not implement this as a method in python and just use the bbox sent by the frontend? So a `map.get_selected_objects()` or so that would read the bounds from `selected_bounds` but then just do that query in Python to get the selected objects? 

I feel like this is almost ready for review -- not sure what else we should do here.

@hanbyul-here would you be able to give this a look and point out anything obvious that you feel should be changed?

There's possibly a few things that could maybe be done nicer?

 - The selection "mode" stuff seems a bit janky: https://github.com/developmentseed/lonboard/pull/417/files#diff-0b5adbfe7b36e4ae2f479291e20152e33e940f7f265162d77f40f6bdb5da7405R287
 - The CSS to position the select control just seems terrible: https://github.com/developmentseed/lonboard/pull/417/files#diff-0b5adbfe7b36e4ae2f479291e20152e33e940f7f265162d77f40f6bdb5da7405R277

Also this was my first time really writing TypeScript so if there's anything there that seems like it's not proper..

One thing to note:

 - Right now, "setting" `map.selected_bounds` in python will not do anything on the JS side - if we want a way for a user to programmatically set the selection in Python, I would rather do this by an explicit method on map like "set_selected_bounds", and treat it similar to how the `flyTo` operation is being handled currently.

kylebarron:
> As an approach, I think I prefer this than trying to send indexes for all the objects down from JS - can we not implement this as a method in python and just use the bbox sent by the frontend? So a `map.get_selected_objects()` or so that would read the bounds from `selected_bounds` but then just do that query in Python to get the selected objects?

I think there are pros and cons to each manner of implementation, but I think that there are some cases where maintaining the indexes on the model itself is valuable. It might also enable more client side use cases where you're syncing the data and the selection with another external widget through JS, avoiding the Python kernel.

And not needing to do another search query on the Python side might be nice for performance as well. And if we're sending up to like 100k uint32s encoded in Parquet to the server, that's a pretty small file. Probably smaller than 50kb?

kylebarron:
Note that a user's selection on the screen is in pixels, and does not always align with an axis-aligned bounding box. E.g. if the map is tilted, the pixel bounding box will accurately choose items within that view because deck.gl uses the actual scene rendering for picking. But that won't be able to be picked from Python because you'd have to serialized the entire view and camera state and implement 3d picking from scratch.

batpad:
@kylebarron based on our conversation, I've made it now so that each layer has a `selected_bounds` property and when a user selects a bbox, it will set that property on each layer on the map.

lonboard/_layer.py
```diff
@@ -183,6 +183,20 @@ def _add_extension_traits(self, extensions: Sequence[BaseExtension]):
     - Default: `False`
     """
 
+    selected_bounds = traitlets.List(
+        trait=traitlets.Float(allow_none=False),
+        allow_none=True,
+        default_value=None,
+        sync=True,
```
kylebarron, line 190: This needs to be `.tag(sync=True)`; I don't think you can pass it in here as you have

batpad:
cc @LanesGood

LanesGood:
@batpad in response to your request for a design review of the bbox selection instructions/icon:

1. is it possible to use the ipyleaflet draw controls? https://ipyleaflet.readthedocs.io/en/latest/controls/draw_control.html
2. If not and this is more of a need to implement custom icons, I'd suggest going with something very recognizable, like the geojson.io controls (which are also used in ipyleaflet): 
image
We could implement this as a button simply in the selection `div` you've added, and I would propose adding a separate "x points selected" text interface element.

vgeorge:
Hey folks, in #567 we fixed the bbox select drawing issues. Items moving forward:

- Merge main into this branch
- Add a toolbar: I believe we need two icons to implement this workflow, one to start a rectangle selection and another to clear an existing selection. This toolbar could also host other features like, exporting image still or [animations](https://github.com/developmentseed/lonboard/issues/405) cc @LanesGood 
- Add a panel to display the number of features selected
- Disable map drag when drawing the rectangle
- Update the cursor type when drawing
- Discuss what is the best approach to select the features; quoting my own comment for the other PR:

> Deck.gl's [pickObjects](https://deck.gl/docs/api-reference/core/deck#pickobjects) doesn't seem to pick all objects in the bbox. It might be because it doesn't pick occluded features. The function [pickMultipleObjects](https://deck.gl/docs/api-reference/core/deck#pickmultipleobjects) seems to be the only method that picks occluded features, but the selection area must be a circle. If we find a way to get the feature IDs, we can implement the [approach](https://github.com/developmentseed/lonboard/pull/417#issuecomment-2020508957) suggested by @kylebarron, but at the moment, the only approach seems to be passing the bbox to the Python code so it can query the features.

vgeorge:
Status Update:

- Implemented XState for state management, replacing the reducer approach. This should simplify handling of side effects and reduce overall complexity
- Enhanced bounding box visuals:
  - Updated style during drawing phase
  - Changed cursor to pointer when drawing
- Integrated dotenv for environment variable management during frontend build. This was necessary for toggling the XState inspector

To Do:

- Add styles to toolbar button to initiate bounding box selection
- Verify correct transmission of bounds to Python core

@kylebarron @batpad following our recent discussion, bounds are being passed to the Python core in [this section](https://github.com/developmentseed/lonboard/blob/0139e831e9508e0b95995cc0104c6168e2199063/src/index.tsx#L156-L161). But I'm unsure how to confirm this in the Jupyter notebook environment.

kylebarron:
> But I'm unsure how to confirm this in the Jupyter notebook environment.

You can add an attribute on the `BaseArrowLayer` that's

```py
class BaseArrowLayer:
    selected_bounds = traitlets.Any().tag(sync=True)
```

Then after a making a selection on the map, you should be able to check `layer.selected_bounds`

vgeorge:
@kylebarron @batpad this is ready for a review. I added a toolbar to handle the different steps of adding a bounding box and connected to the layer attributes `selected_bounds` property. Colors and icons are not super great, but I want to improve them in a separated ticket. Here is a screen capture of the current workflow:

bbox-select

lonboard/_layer.py
```diff
@@ -262,6 +278,14 @@ class BaseArrowLayer(BaseLayer):
...
 
     table: traitlets.TraitType
+    selected_bounds: traitlets.Tuple(
+        traitlets.Float(),
+        traitlets.Float(),
+        traitlets.Float(),
+        traitlets.Float(),
+        allow_none=True,
+        default_value=None,
+    ).tag(sync=True)
```
kylebarron, line 288: This doesn't need to be on the BaseArrowLayer as well. It will inherit from `BaseLayer`.

src/index.tsx
```diff
@@ -163,16 +240,28 @@ function App() {
...
         useDevicePixels={isDefined(useDevicePixels) ? useDevicePixels : true}
         // https://deck.gl/docs/api-reference/core/deck#_typedarraymanagerprops
         _typedArrayManagerProps={{
           overAlloc: 1,
           poolSize: 0,
         }}
+        onLoad={() => {
+          actorRef.send({
+            type: "Deck.gl was loaded",
+          });
```
kylebarron, line 263: What does this do?

kylebarron (commented):
This looks good at a glance to me, though I'd love @batpad to check it out as well if he has time!

kylebarron (commented):
This looks good at a glance to me, though I'd love @batpad to check it out as well if he has time!

src/index.tsx
vgeorge: This is an action in the XState machine. It was added to signal that the Deck.gl component is ready, but in the current workflow it is not necessary and I removed it.

batpad (approved):

vgeorge:
@kylebarron @batpad thanks for the review, I'll merge this now. Sanjay and I had a quick chat about the XState implementation and we agreed some documentation about it would be helpful. I'll open a ticket to track it.

Issue #425, opened 2024-03-20
Support `__arrow_c_array__` in viz()


kylebarron:
It would be nice to be able to visualize any array. Note that this should be before `__geo_interface__` in the conversion steps.

You might want to do something like the following to ensure the field metadata isn't lost if extension types aren't installed.

```py
if hasattr(obj, "__arrow_c_array__"):
    schema, _ = obj.__arrow_c_array__()

    class SchemaHolder:
        def __init__(self, capsule) -> None:
            self.capsule = capsule

        def __arrow_c_schema__(self):
            return self.capsule
        
    pyarrow_field = pa.field(SchemaHolder(schema))
    pyarrow_array = pa.array(obj)
```

PR #427, opened 2024-03-20
Support geoarrow array input into `viz()`

kylebarron:
Closes https://github.com/developmentseed/lonboard/issues/425, blocked on https://github.com/apache/arrow/issues/38010#issuecomment-2010601912. The main issue is that we need a reliable way to maintain the geoarrow extension metadata through FFI. The easiest way would be if `pa.field()` were able to support `__arrow_c_schema__` input. Or alternatively, one option is to have a context manager of sorts to register global pyarrow geoarrow extension arrays, and then deregister them after use.


kylebarron:
As noted in https://github.com/apache/arrow/issues/38010#issuecomment-2018174213, for now we can use `pa.Field._import_from_c_capsule`

kylebarron:
Once pyarrow v16 is released, we should make sure we can visualize chunked geometry arrays as well. But for now, we can only import an Arrow stream to a table, and that stream must have struct batches.

PR #444, opened 2024-03-30
sync view state, move over code from #113, refs #112

batpad:
Created a new PR toward #112, essentially copying over the code from #113, but applying it against `main`.

@kylebarron this seems to work!! It does seem to make the map a bit slow for me, and we should investigate that, but in principle, this for me works as expected.

On the javascript, I can move and zoom the map around and then in python call:

    m._view_state

Where `m` is a `Map` instance, and I get the current `viewState` as it is on the map.

Conversely, to modify the view state on the map, in Python I can do something like:

    new_state = m._view_state.copy()
    new_state['zoom'] = 8
    m._view_state = new_state

The above code works to set the `zoom` on the map to 8.

I think it will still be great to get eyes on this from React folks, and maybe think about if we could upstream the `useModelStateDebounced` method or so, but in principle, this seems to work!

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -15,7 +15,6 @@ import { Message } from "./types.js";
 import { flyTo } from "./actions/fly-to.js";
 import { useModelStateDebounced } from "./state";
 
-
 await initParquetWasm();
 
 const DEFAULT_INITIAL_VIEW_STATE = {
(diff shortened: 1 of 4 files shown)


src/index.tsx
```diff
@@ -165,6 +170,15 @@ function App() {
           overAlloc: 1,
           poolSize: 0,
         }}
+        viewState={
+          Object.keys(viewState).length === 0
+            ? DEFAULT_INITIAL_VIEW_STATE
+            : viewState
+        }
```
kylebarron, line 177: I'd guess we don't want to check this on every render

kylebarron:
Closing in favor of https://github.com/developmentseed/lonboard/pull/448, which is based off this pr

PR #448, opened 2024-04-01
Sync view state between JS and Python

kylebarron:
Closes https://github.com/developmentseed/lonboard/issues/112

In https://github.com/developmentseed/lonboard/pull/444 @batpad mentioned 

> It does seem to make the map a bit slow for me, and we should investigate that, but in principle, this for me works as expected

While playing around with that PR, I also saw some view state lagging at low zooms with lots of data rendered. In this screencast I'm smoothly zooming in, but the rendering is very jagged, and sometimes zooms back out slightly. **This screencast is of the previous PR**

https://github.com/developmentseed/lonboard/assets/15164633/bbe0a168-b76e-4563-bc30-828d648210db

My hypothesis of what's happening is that when rendering lots of data there's a slight lag for deck to call the `onViewStateChange` handler. And so deck will occasionally call the handler with "old" data and that's where you sometimes get the "snap back" in the scrolling.

I think this is a really bad UX, and it seems not to happen when deck manages the view state internally. So I think potentially the best UX and performance solution is to decouple the view state from the model, while still syncing from deck to Python in the debounced model handler.

This PR tries to implement this.

### DIFF
--- a/src/state.ts
+++ b/src/state.ts
@@ -3,7 +3,6 @@ import { useModel } from "@anywidget/react";
 import { debounce } from "./util";
 
 const debouncedModelSaveViewState = debounce((model) => {
-  console.log("DEBOUNCED");
   const viewState = model.get("_view_state");
 
   // transitionInterpolator is sometimes a key in the view state while panning
@@ -31,12 +30,9 @@ export function useModelStateDebounced<T>(
   let [value, setValue] = React.useState(model.get(key));
   React.useEffect(() => {
     let callback = () => {
-      console.log("callback");
-      console.log(model.get(key));
       setValue(model.get(key));
     };
     model.on(`change:${key}`, callback);
-    console.log(`model on change view state`);
     return () => model.off(`change:${key}`, callback);
   }, [model, key]);
   return [
--- a/src/state.ts
+++ b/src/state.ts
@@ -1,8 +1,9 @@
 import * as React from "react";
 import { useModel } from "@anywidget/react";
+import type { AnyModel } from "@anywidget/types";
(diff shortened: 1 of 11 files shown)


src/state.ts
```diff
@@ -0,0 +1,48 @@
...
+  // transition state it expects, because hopefully the transition will have
+  // finished in the 300ms that the user has stopped panning.
+  if ("transitionInterpolator" in viewState) {
+    console.debug("Deleting transitionInterpolator!");
+    delete viewState.transitionInterpolator;
+    model.set("_view_state", viewState);
+  }
+
+  model.save_changes();
+}, 300);
```
kylebarron, line 24: Note that this is very tied to view state, so maybe we shouldn't name the below function  `useModelStateDebounced`

kylebarron:
I started prototyping using an evented dict to hold the view state, but I'm going to punt on that for now. Here was the code I attempted with:

```py

def serialize_view_state(data: EventedDict, obj):
    return json.dumps(data._dict)


class ViewStateTrait(FixedErrorTraitType):
    default_value = EventedDict()

    def __init__(
        self: TraitType,
        *args,
        **kwargs: Any,
    ) -> None:
        super().__init__(*args, **kwargs)

        self.tag(sync=True, to_json=serialize_view_state)

    def validate(self, obj, value):
        # TODO: validate view state keys
        evented_dict = EventedDict(value)

        @evented_dict.events.connect
        def send_state():
            obj.send_state("view_state")

        return evented_dict
```

batpad (approved):

Issue #204, opened 2023-11-06
Docs: Security/Authentication


kylebarron:
Right now, all data is fetched from Python directly, and then the data moves from Python to JS. This means that JS doesn't need any security credentials to know how to fetch the data. This may change in the future with tile layers that fetch data directly from JS

PR #450, opened 2024-04-04
Internal architecture documentation

kylebarron:
Closes https://github.com/developmentseed/lonboard/issues/204


docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+Each layer has parameters that map to deck.gl accessors and render properties. Each of these has validation on the Python side to catch invalid input early. Accessors are modeled with ["trait" objects](https://developmentseed.org/lonboard/latest/api/traits/). Scalar accessors are stored as a scalar int/string/float. Function accessors are evaluated in Python and stored as a `pyarrow.Array`.
+
+A variety of [layer extensions](https://developmentseed.org/lonboard/latest/api/layer-extensions/) are supported, which map to deck.gl's upstream layer extensions. These layer extensions dynamically add more parameters to the layers on which they're assigned.
+
+
+### Data loading
+
+While pydeck allows users to pass in URLs to remote files, at the moment this is not possible in Lonboard. With only a few exceptions (e.g. the `BitmapLayer` and `BitmapTileLayer`), _only_ data originating in Python is supported in Lonboard. This reduction in scope vastly simplifies Lonboard's internals while not providing much of a hurdle for Lonboard's core audience.
```
felixpalmer, line 47: > only a few exceptions

This begs the question, why allow just these exceptions? They show that it is possible to pass through a URL and let deck "do the rest". Is the idea to add more in the future, or is there something special about `BitmapLayer`?

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
+# Internal Architecture
+
+> Last edited April 2024.
+
+This document will outline Lonboard's internal architecture, and hopefully should be useful for potential contributors.
+
+## Pydeck
+
+Before we consider Lonboard, we should briefly cover how Pydeck manages data transit.
```
felixpalmer, line 9: > how Pydeck manages data transit.

Definitely agree that serializing a Python dataframe to a string to pass via pydeck doesn't scale. Thus the easiest way to work with pydeck is via a URL in the data prop.

However Pydeck does also support a [Binary Data Transfer](https://deckgl.readthedocs.io/en/latest/binary_transfer.html) which uses the same idea as lonboard to transfer the binary payload via websockets. 

What is missing from pydeck (and arguably deck.gl by extension) is a way render the binary data on the other end. However, you have implemented a collection of [GeoArrow-compatible layers
](https://github.com/geoarrow/deck.gl-layers), is there any reason these couldn't be used via pydeck?

I did a quick test to compare pydeck vs lonboard rendering 3M points and while lonboard is faster, it certainly isn't orders of magnitude:

_videos trimmed to show time from notebook cell refresh to first paint_

https://github.com/developmentseed/lonboard/assets/453755/ebe7abb0-a636-4a39-8d34-39b04e351c62

https://github.com/developmentseed/lonboard/assets/453755/7931c874-c02a-454f-803a-9c62d3127f2b

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+<!-- It's impossible to consider Lonboard's architecture in isolation because many of Lonboard's architectural decisions were chosen in explicit contrast to pydeck. Lonboard and pydeck both bind to deck.gl, but have significantly different APIs and characteristics. -->
```
felixpalmer, line 11: > both bind to deck.gl

I would actually argue that lonboard modifies the internals of deck layers to such an extent that it is more accurate to think of it as a Python GeoArrow rendering toolkit that is built on top of deck.gl, rather than a binding

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+These `Layer` classes subclass from `ipywidgets.Widget` but are _not_ associated with any JavaScript of their own. But by being widgets themselves, their data is synced with JavaScript and Python data changes are propagated as events to the core `Map` object.
+
+Each layer has parameters that map to deck.gl accessors and render properties. Each of these has validation on the Python side to catch invalid input early. Accessors are modeled with ["trait" objects](https://developmentseed.org/lonboard/latest/api/traits/). Scalar accessors are stored as a scalar int/string/float. Function accessors are evaluated in Python and stored as a `pyarrow.Array`.
```
felixpalmer, line 40: > Function accessors are evaluated in Python

This is analogous to [supplying attributes directly](https://deck.gl/docs/developer-guide/performance#supply-attributes-directly) in JavaScript, correct? In that sense they are not so much "function accessors" but rather pre-computed attributes. IMHO the name "accessor" doesn't work well here, if functions are not supported

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+#### Simplicity of code
+
+Supporting only a single data provenance and format makes the internals of Lonboard much more maintainable. Lonboard's JavaScript bindings only need to include a single data parser and can optimize performance for Parquet and Arrow data.
+
+#### Parquet
+
+Parquet was chosen as the underlying data transmission format for its excellent compression, fast read and write speeds, compatibility with GeoArrow and, in turn, compatibility with deck.gl's binary API.
+
+I tested with data from the speed test example (3 million points). Saving GeoArrow to Parquet was 135x faster than saving a GeoDataFrame to GeoJSON. The resulting file was 26x smaller. And parsing the Parquet to Arrow on the frontend was 5.6x faster than `JSON.parse` (including moving the Arrow from WebAssembly to JS).
```
felixpalmer, line 59: See comment above on speed comparison.

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+#### Bootstrapping GeoArrow
+
+For performance reasons, Lonboard is very focused on using GeoArrow throughout the pipeline in both Python and JavaScript. At the time when Lonboard was started, there were no user-friendly tools for converting data to GeoArrow in client-side JavaScript. By supporting only data from Python and only Parquet data, we could focus on making data conversions from GeoPandas to GeoArrow in Python efficient and stable. This enables GeoArrow to be "just an implementation detail" and out of any primary public APIs while also working to bootstrap a GeoArrow ecosystem in Python and JavaScript.
+
+#### Authentication
+
+Since all data is loaded via Python over the Jupyter-provided WebSocket, Lonboard doesn't need to implement any sort of authentication itself. The user is able to use whatever credentials they're familiar with, say for S3, to access data. Those credentials never need to be passed to the browser.
+
+In contrast, having the JavaScript side load data directly from a URL requires some way for users to specify authorization headers or API keys for access to private datasets.
```
felixpalmer, line 72: Any thoughts on how this would work for different basemaps, like Google/Mapbox which require API keys?

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+getRadius: (object) => object.properties.valuePerSqm
+```
+
+on the JavaScript side.
+
+I find this syntax to be horrible UX. The user is no longer writing Python, they need to learn a new DSL just for this.
+
+RGB Colors for 3 million points take up 9MiB in memory (or 12MiB for RGBA). In the example of speed test data, 12MiB of RGBA data compressed down to 4MiB for transit. Float32 radii values for 3 million points take up 12MiB in memory and 9MiB compressed.
+
+Even for users on remote Python environments, this takes no more than a couple seconds to download, and then can render almost instantly. For non-point datasets with a fewer number of geometries, the byte size of accessors are even smaller relative to the geometry buffers.
```
felixpalmer, line 108: > a couple seconds to download, and then can render almost instantly

For a user, this is just "how long it takes my viz to update", they don't care if the render is instant if they have to wait seconds for the data. Contrast to applying a color map in JavaScript, which would feel instant (just playing devil's advocate here, I know many users will want to apply their color maps in python, even if it is slow)

docs/how-it-works/internals.md
kylebarron: > However Pydeck does also support a [Binary Data Transfer](https://deckgl.readthedocs.io/en/latest/binary_transfer.html) which uses the same idea as lonboard to transfer the binary payload via websockets.

Yes, but with a few caveats:

- Pydeck supports only numeric binary arrays. So no strings nor nested structs or lists. This _includes_ any attribute columns, as far as I can tell. So if you want a tooltip with a string or struct column, you can't use binary transport.

    This also means it can't be used with any layer types other than points. Loaders.gl does have a binary data representation for other geometry types, but you'd have to manage conversion from Python geometries to that representation yourself. Which essentially re-implements GeoArrow.
- It's opt-in, and the user needs to know not only that the `use_binary_transport` param exists, but also its caveats, and which of their datasets can take advantage of the option.
- Pydeck's binary transport isn't compressed at all. I assume you tested in a local Python environment, where Python is on the same machine as the browser. Some of Devseed's clients work on remote JupyterLab (JupyterHub) environments, and so effective compression is a must. For large data sizes, gzip can be unacceptably slow (both to compress and to decompress, in my opinion). ZSTD is the best compression available for this type of data, but then you need a Wasm dependency on the frontend.
- Lonboard automatically manages data chunking. It looks like pydeck serializes columns individually but does not do any chunking. So for 3 million points it sends a single 48MB buffer for the `getPosition` accessor, which from a remote server can easily crash the websocket connection.

> What is missing from pydeck (and arguably deck.gl by extension) is a way render the binary data on the other end. However, you have implemented a collection of [GeoArrow-compatible layers ](https://github.com/geoarrow/deck.gl-layers), is there any reason these couldn't be used via pydeck?

No reason it couldn't be used, but you'd need to create Arrow tables. And pydeck looks like it's already using [binary data rendering](https://github.com/visgl/deck.gl/blob/33f8d2dd4ab98963f25c6e6914107491d89de143/bindings/pydeck/pydeck/data_utils/binary_transfer.py#L46-L49). It's unclear what pydeck would gain from using these layers except for non-point support. But then you'd need to either use Arrow over the network or manage your own buffer transport for nested types.

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+This document will outline Lonboard's internal architecture, and hopefully should be useful for potential contributors.
+
+## Pydeck
+
+Before we consider Lonboard, we should briefly cover how Pydeck manages data transit.
+
+<!-- It's impossible to consider Lonboard's architecture in isolation because many of Lonboard's architectural decisions were chosen in explicit contrast to pydeck. Lonboard and pydeck both bind to deck.gl, but have significantly different APIs and characteristics. -->
+
+Pydeck uses the loaders.gl ecosystem natively supported in deck.gl. This means that pydeck passes the user input directly into the `data` prop of any given layer. When `data` is a string, it's interpreted as a URL, passed into loaders.gl machinery to parse the given data format to a list of GeoJSON features. In order to do data-driven styling, users pass in strings like `"@@=properties.valuePerSqm"` to describe the JavaScript callback that should be applied to each GeoJSON feature.
```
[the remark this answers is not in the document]
kylebarron: > > GeoJSON features
> 
> Only true for `GeoJsonLayer`, in general layers can work with an arbitrary collection of data

s/GeoJSON features/JSON objects

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+In my opinion, this presents performance pitfalls and gives Pydeck a stifling amount of flexibility. The Python API isn't able to do much of anything because the user input could be _anything_ and users end up writing a sort of JavaScript in Python strings.
```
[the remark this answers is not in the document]
kylebarron: > > The Python API isn't able to do much of anything
> 
> Not entirely accurate, there is nothing stopping the user transforming the data and passing it in the data prop. The accessor then just references the Python-computed property by key. You can certainly point out perf weaknesses here, but it's not like it is impossible to apply transformations in Python

I'm referring to pydeck's own Python code. Of course the user can do whatever they want before passing data into the Pydeck API. But Pydeck is not able to do anything smart with the user input because there's no constraint on the shape of data the user can pass in. By constraining to GeoPandas GeoDataFrames and Arrow tables, Lonboard's _own code_ is able to do a lot of smart stuff for the user, without them needing to be aware/learn anything.

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+- Wants fast exploratory data analysis.
+- Already working with data in Python.
+- A reasonably fast internet connection on a desktop computer.
+
+### Python classes
+
+The `Map` class is the primary and only widget associated with JavaScript that renders anything. `Map` subclasses from `anywidget.AnyWidget`, which dynamically fetches Lonboard's JavaScript ESM bundle. The `Map` class emulates deck.gl's `Deck` class. It synchronizes map state and can be passed a sequence of `Layer` objects.
+
+`Layer` classes are designed to map to each underlying deck.gl layer. Most `Layer` classes subclass from `BaseArrowLayer`, which stores a `pyarrow.Table` in the dataclass. This data object gets serialized to Parquet and rendered in a layer provided by `@geoarrow/deck.gl-layers` as described later on. The primary API for each of these Arrow-based layers is `from_geopandas`, which converts a `GeoDataFrame` to a `pyarrow.Table` with a [GeoArrow](https://geoarrow.org/)-represented geometry column. A few layers are not Arrow-based and do not subclass from `BaseArrowLayer`: primarily `BitmapLayer` and `BitmapTileLayer`. These pass user data input directly into a core deck.gl layer.
```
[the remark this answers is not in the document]
kylebarron: > > provided by `@geoarrow/deck.gl-layers`
> 
> I think the layers you've implemented here are actually some of the most valuable contributions, allowing a zero-copy way to render GeoArrow data. These have value in and of themselves and could be used in other contexts - e.g. a web app wanting to render GeoArrow.

I'm certainly interested in better integration on the JS side between deck and GeoArrow. I think those layers could fit into a `@deck.gl/arrow` module (assuming that an Arrow JS dependency wouldn't fit into any other submodule).

We've started to prototype web apps that use GeoArrow internally and `@geoarrow/deck.gl-layers` for rendering.

docs/how-it-works/internals.md
kylebarron: > > Function accessors are evaluated in Python
> 
> This is analogous to [supplying attributes directly](https://deck.gl/docs/developer-guide/performance#supply-attributes-directly) in JavaScript, correct? In that sense they are not so much "function accessors" but rather pre-computed attributes. IMHO the name "accessor" doesn't work well here, if functions are not supported

I'm using the name "accessor" to provide consistency with the meaning of the JS API. In this context I'm using "accessor" to refer to any property that starts with the `get` prefix.

I've considered whether to emulate the functional API on the Python side. Where the user could pass 

```py
ScatterplotLayer(
    get_radius=lambda object: object["value"] * 1000
)
```

but I'm hesistant to do this because this requires a slow Python for loop. If we nudge the user to do 

```py
ScatterplotLayer(
    get_radius=dataframe["value"] * 1000
)
```
then numpy evaluates that in C, we don't have to rebuild the buffers, and it'll be a lot faster without the user needing to understand the difference between those two.

docs/how-it-works/internals.md
kylebarron: As mentioned above, Lonboard is able to provide a good user experience with good performance because of the constraints it enforces on user input. This allows Lonboard to apply performance improvements under the hood without the user needing to learn anything special.

If we added a `GeoJsonLayer` which sent data over the wire as GeoJSON and used deck's `GeoJsonLayer`, then the user would have to understand the difference between that layer and the other layers, and we'd get issues about this layer being much smaller.

`BitmapLayer` is special because it doesn't have a direct GeoArrow counterpart.

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+```js
+getRadius: (object) => object.properties.valuePerSqm
+```
+
+on the JavaScript side.
+
+I find this syntax to be horrible UX. The user is no longer writing Python, they need to learn a new DSL just for this.
+
+RGB Colors for 3 million points take up 9MiB in memory (or 12MiB for RGBA). In the example of speed test data, 12MiB of RGBA data compressed down to 4MiB for transit. Float32 radii values for 3 million points take up 12MiB in memory and 9MiB compressed.
```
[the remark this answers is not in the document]
kylebarron: It's not a goal of Lonboard to always strictly minimize data transfer. Rather, Lonboard's ethos is: if we can make the serialization _efficient enough_, then we _can_ download some data twice. Python users don't know and don't want to know JS, and don't want to think about what data transformations happen on the frontend. Letting all data transformations happen in the user's Python environment provides a level of certainty and transparency that doesn't happen for client-side transformations. A user can apply colormaps using whatever method they're using to in Python already.

It's fine to download this amount of data twice, which even on a remote connection likely only takes a couple seconds

docs/how-it-works/internals.md
kylebarron: For local Python environments, applying the colormap on the Python side is just as fast as doing it in JS. For remote environments, I believe the "understandability" of Python code is preferred to the user somehow serializing a colormap to JS and applying it there. I think the benefit of having Lonboard be "just Python" is worth waiting 1-2 seconds for a new colormap when you have 3 million rows.

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+which would be expanded into
+
+```js
+getRadius: (object) => object.properties.valuePerSqm
+```
+
+on the JavaScript side.
+
+I find this syntax to be horrible UX. The user is no longer writing Python, they need to learn a new DSL just for this.
```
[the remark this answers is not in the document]
kylebarron: > And if the pydeck binary transport is used, the API is close to lonboard, just rather than passing in the column, the column name in the dataframe is referenced:
> 
> ```python
> pdk.Layer(
>     "ScatterplotLayer",
>     df,
>     get_position="position",
>     ...
> )
> ```

In this case, updating an accessor from Python would require sending the entire dataset from Python to JS again.

docs/how-it-works/internals.md
```diff
@@ -0,0 +1,120 @@
...
+
+With Lonboard's widget architecture, any accessors other than the geometry are stored separately to the main Arrow table. This means that all accessors and rendering properties can be updated in isolation. The core table does not need to be synced again when a property is updated.
+
+#### Uses familiar objects
+
+Accessors are defined in terms of Numpy arrays, Pandas Series or PyArrow Arrays, which are familiar to Python data scientists.
+
+#### Full access to Python
+
+Because we serialize _data_ to JavaScript instead of a function in a custom DSL, the user has the full spectrum of Python available to them. They can use one of matplotlib's many [normalization helpers](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.Normalize.html) before [applying a colormap](https://developmentseed.org/lonboard/latest/api/colormap/#lonboard.colormap.apply_continuous_cmap). Or they can use apply some ML model and use its outputs for point radii.
```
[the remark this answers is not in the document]
kylebarron: In the special case of opt-in binary transport of point geometries with no string columns, the data transfer can be efficient enough to let users apply data updates in pure Python. But for other cases, when sending the data to JS is so expensive, the user is incentivized into doing data operations in JS strings.

PR #458, opened 2024-04-09
Try to fix docs ci

kylebarron:


kylebarron:
Fixed by https://github.com/developmentseed/lonboard/pull/460 instead

Issue #457, opened 2024-04-08
import parquet-wasm from local bundle rather than CDN


cornhundred:
Hi, is it possible to import `parquet-wasm` using esbuild rather than a CDN? I see that `parquet-wasm` is a dependency in the package.json

https://github.com/developmentseed/lonboard/blob/c8cf2e1862fa7cd84af196039b21a97a5f00fd8f/package.json#L5-L20

but lonboard is also referencing a CDN here:

https://github.com/developmentseed/lonboard/blob/c8cf2e1862fa7cd84af196039b21a97a5f00fd8f/src/parquet.ts#L2-L19

I ask because I'm trying to test out lonboard on a service that prevents me from making front-end requests to CDNs (I would have to request the URL be added to an allowlist, but I want to see if I can avoid this by bundling the code with esbuild but I was running into errors when I tried to use the bundled version - https://github.com/kylebarron/parquet-wasm/issues/488).

kylebarron:
I agree, this change would be nice. But it's unclear to me how to get a reference to the local URL within Jupyter to the current bundle. I asked in the anywidget discord

cornhundred:
Thanks @kylebarron, I was able to find a workaround that might be relevant here https://github.com/kylebarron/parquet-wasm/issues/488#issuecomment-2047904050

kylebarron:
Trevor responded on slack that serializing the wasm on the model is probably the best approach

prusswan:
Until #461 gets merged, the easiest way (imho) is to let user specify a different cdn url (e.g. http://localhost:8888/files/). For now, we need to resort to modifying the js bundle.

kylebarron:
What's your use case where you can't use the default CDN?

prusswan:
Offline or standalone usage

kylebarron:
But offline you won't be able to see the base map; does that not matter for you?

prusswan:
> But offline you won't be able to see the base map; does that not matter for you?

Usually I have a copy of the tiles for performance/practical reasons - it is faster than getting over the internet, while in other cases I am generating them anyway. If BitmapTileLayer (which already works with locally hosted Jupyter file urls) can be expanded to support/include pmtiles it would be perfect (sounds like it is in the works with #494)

PR #461, opened 2024-04-10
Send WebAssembly binary over Jupyter WebSocket

kylebarron:
[We use Parquet](https://developmentseed.org/lonboard/latest/how-it-works/internals/#parquet) as the internal format for data transfer. For reasons linked, Parquet is great. But to read Parquet on the client, we need to use a Wasm-based Parquet reader like my own https://github.com/kylebarron/parquet-wasm. Wasm-based libraries need a sidecar binary `.wasm` file, which is usually distributed separately.

We currently fetch this file from CDN, but for environments with a strong outbound firewall, a CDN may not be allowed. See https://github.com/developmentseed/lonboard/issues/457. To get around this, we serialize the gzipped Wasm content on the Anywidget model itself. We then decompress it on the client and pass it into Parquet-wasm's initializer.

Closes https://github.com/developmentseed/lonboard/issues/457

### DIFF
--- a/lonboard/_map.py
+++ b/lonboard/_map.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import sys
+from functools import lru_cache
 from pathlib import Path
 from typing import IO, TYPE_CHECKING, Optional, Sequence, TextIO, Union
 
@@ -44,6 +45,13 @@ _HTML_TEMPLATE = """<!DOCTYPE html>
 </body>
 </html>
 """
+
+
+@lru_cache
+def _load_parquet_wasm_binary() -> bytes:
+    """Load the gzipped parquet-wasm binary blob"""
+    with open(bundler_output_dir / "arrow2_bg.wasm.gz", "rb") as f:
+        return f.read()
 
 
 class Map(BaseAnyWidget):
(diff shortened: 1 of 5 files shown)


manzt:
one suggestion, you can use some indirection to create a model with just the static contents so that it is hoisted from each model instance:

```py
import ipywidgets
import anywidget
import traitlets


class StaticAsset(ipywidgets.Widget):
    contents = traitlets.Any().tag(sync=True)

asset = StaticAsset(contents=b"hello, world")
    
class Widget(anywidget.AnyWidget):
    _esm = """
    async function load_asset(model, name) {
        let model_id = model.get(name).slice("IPY_MODEL_".length);
        let asset_model = await model.widget_manager.get_model(model_id);
        return asset_model.get("contents");
    }
    async function render({ model, el }) {
        let asset = await load_asset(model, "asset")
        el.innerText = new TextDecoder().decode(asset)
    }
    export default { render }
    """
    asset = traitlets.Any(asset).tag(sync=True, **ipywidgets.widget_serialization)
    

Widget()
```

Each widget instance will just have `IPY_MODEL_xxxx`, and reuse the asset in the front end. I am working on making anywidget hoist `_esm` and `_css` assets this way to avoid duplication in the HTML.

manzt:
It should be noted this is just a way to get the static assets into the front end, but derived objects (e.g., initialized parquet) would need to be cached somewhere. Probably easiest to make some global for now, but would be more elegant if anywidget could provide a more ideomatic API.

Issue #484, opened 2024-04-23
PathStyleExtension in lonboard


shriv:
As noted in #473 my use case is plotting link volumes on NZ roads. One "feature" of the road geometries (which come from OpenStreetMap) is that lane geometries lie on top of each other. This doesn't look good in the visualisation so we want to split paths with some offset. One fix seems to be the `deck.gl` `PathStyleExtension` with the `offset:true` option. I was wondering if there is a way to include this in a `lonboard` map even though it is not available in the [current set](https://developmentseed.org/lonboard/latest/api/layer-extensions/)? 

```
new deck.PathLayer({
      widthMinPixels: 1,
      capRounded: true,
      data: pathData,
      getPath: (d) => d.path,
      getColor: (d) => d.color,
      getWidth: (d) => d.width,
      getOffset: -0.5,
      extensions: [
        new deck.PathStyleExtension({
          offset: true
        })
      ],
      parameters: {
        depthTest: false
      }
    })
```

kylebarron:
Yes it definitely makes sense to add that to lonboard. 

Do you think you'd be able to try to implement it? It should largely be straightforward if you look at how the other extensions are implemented

shriv:
Yep, I can definitely give it a go :-)

kylebarron:
Closed by https://github.com/developmentseed/lonboard/pull/487

PR #487, opened 2024-04-25
Adding PathStyleExtension code

shriv:
I've done a very basic addition of `PathStyleExtension` to address #484. One key TODO is checking the accessor for `get_dash_array`. I've noted it in the class doc. 

This is my first PR so I'm not sure if I've missed anything. Let me know! 😄

### DIFF
--- a/lonboard/layer_extension.py
+++ b/lonboard/layer_extension.py
@@ -1,7 +1,7 @@
 import traitlets
 
 from lonboard._base import BaseExtension
-from lonboard.traits import FloatAccessor, GetFilterValueAccessor, PointAccessor
+from lonboard.traits import FloatAccessor, GetFilterValueAccessor, PointAccessor, DashArrayAccessor
 
 
 class BrushingExtension(BaseExtension):
@@ -375,7 +375,7 @@ class PathStyleExtension(BaseExtension):
     _extension_type = traitlets.Unicode("path-style").tag(sync=True)
 
     _layer_traits = {
-        "get_dash_array": FloatAccessor(None, allow_none=True),
+        "get_dash_array": DashArrayAccessor(None, allow_none=True),
         "dash_justified": traitlets.Bool(False).tag(sync=True),
         "get_offset": FloatAccessor(None, allow_none=True),
         "dash_gap_pickle": traitlets.Bool(False).tag(sync=True),
--- a/lonboard/layer_extension.py
+++ b/lonboard/layer_extension.py
@@ -1,8 +1,12 @@
 import traitlets
 
 from lonboard._base import BaseExtension
-from lonboard.traits import FloatAccessor, GetFilterValueAccessor, \
-    PointAccessor, DashArrayAccessor
+from lonboard.traits import (
+    DashArrayAccessor,
+    FloatAccessor,
+    GetFilterValueAccessor,
+    PointAccessor,
+)
 
 
 class BrushingExtension(BaseExtension):
@@ -148,8 +152,8 @@ class DataFilterExtension(BaseExtension):
     )
     ```
 
-    The `DataFilterExtension` allows filtering on 1 to 4 attributes at the same time. 
-    So if you have four numeric columns of interest, you can filter on the 
+    The `DataFilterExtension` allows filtering on 1 to 4 attributes at the same time.
+    So if you have four numeric columns of interest, you can filter on the
     intersection of all of them.
 
     For easy visualization, we suggest connecting the `DataFilterExtension` to an
@@ -315,7 +319,7 @@ class PathStyleExtension(BaseExtension):
     # Example
 
     Deck.gl example https://deck.gl/docs/api-reference/extensions/path-style-extension
-    TODO Add Lonboard example. 
+    TODO Add Lonboard example.
 
     # Layer Properties
 
@@ -323,7 +327,7 @@ class PathStyleExtension(BaseExtension):
     it is included:
 
     ## `get_dash_array`
-    
+
     Must be specified if the dash option is enabled. The dash array to draw each path
     with [dashSize, gapSize] relative to the width of the path.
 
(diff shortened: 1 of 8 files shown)


mkdocs.yml
```diff
@@ -68,6 +68,7 @@ nav:
           - api/layer-extensions/brushing-extension.md
           - api/layer-extensions/collision-filter-extension.md
           - api/layer-extensions/data-filter-extension.md
+          - api/layer-extensions/path-style-extension.md
```
kylebarron, line 71: This file doesn't get auto-generated. We have to create a new markdown at that location. You can look at `data-filter-extension.md` for the syntax.

src/model/extension.ts
```diff
@@ -135,6 +136,45 @@ export class DataFilterExtension extends BaseExtensionModel {
...
+  ) {
+    super(model, updateStateCallback);
+
+    const dash = this.model.get("dash");
+    const highPrecisionDash = this.model.get("high_precision_dash");
+    const offset = this.model.get("offset");
+    this.extensionInstance = new _PathStyleExtension({ dash });
+
+    // Properties added by the extension onto the layer
+    layerModel.initRegularAttribute("get_dash_array", "getDashArray");
```
kylebarron, line 157: Anything that starts with `get` is an "accessor" not an "attribute", so this should be `initVectorizedAccessor` instead. (If you could group the `initVectorizedAccessor` lines at the bottom that would be great)

src/model/extension.ts
```diff
@@ -135,6 +136,45 @@ export class DataFilterExtension extends BaseExtensionModel {
...
+    layerModel.initRegularAttribute("dash_justified", "dashJustified");
+    layerModel.initRegularAttribute("get_offset", "getOffset");
```
kylebarron, line 159: ditto

src/model/extension.ts
```diff
@@ -166,8 +206,16 @@ export async function initializeExtension(
...
+    
+    case PathStyleExtension.extensionType:
+      extensionModel = new PathStyleExtension(
+        model,
+        layerModel,
+        updateStateCallback,
+      );
+      break;
+    
+        default:
```
kylebarron, line 218: This should be dedented (that should also help fix CI). You can run `yarn install && yarn run fmt` from the root of the directory to run code formatting on the javascript code.

src/model/extension.ts
```diff
@@ -135,6 +136,45 @@ export class DataFilterExtension extends BaseExtensionModel {
...
+    layerModel.initVectorizedAccessor("get_filter_value", "getFilterValue");
+
+    // Update the layer model with the list of the JS property names added by
+    // this extension
+    layerModel.extensionLayerPropertyNames = [
+      ...layerModel.extensionLayerPropertyNames,
+      "getDashArray",
+      "dashJustified",
+      "getOffset",
+      "dashGapPickle",
```
kylebarron, line 173: Can you sort these lines alphabetically? Also you're missing `getFilterValue`

src/model/extension.ts
kylebarron: And by "accessor" I mean that the user can pass in either a scalar to apply to _every_ row or a column of values where each row applies to the geometry at the same index.

src/model/extension.ts
```diff
@@ -135,6 +136,45 @@ export class DataFilterExtension extends BaseExtensionModel {
...
+    model: WidgetModel,
+    layerModel: BaseLayerModel,
+    updateStateCallback: () => void,
+  ) {
+    super(model, updateStateCallback);
+
+    const dash = this.model.get("dash");
+    const highPrecisionDash = this.model.get("high_precision_dash");
+    const offset = this.model.get("offset");
+    this.extensionInstance = new _PathStyleExtension({ dash });
```
kylebarron, line 154: I think you want to pass in `highPrecisionDash` and `offset` as well. We should only pass in defined parameters though, so you should probably use an `isDefined` cantonation like

```
new _PathStyleExtension({
      ...(isDefined(dash) && { dash }),
      ...(isDefined(highPrecisionDash) && { highPrecisionDash }),
      ...(isDefined(offset) && { offset }),
```
in this case, if any of those values are `undefined`, then it won't get passed in to the constructor

src/model/extension.ts
```diff
@@ -135,6 +136,45 @@ export class DataFilterExtension extends BaseExtensionModel {
...
+
+    // Properties added by the extension onto the layer
+    layerModel.initRegularAttribute("get_dash_array", "getDashArray");
+    layerModel.initRegularAttribute("dash_justified", "dashJustified");
+    layerModel.initRegularAttribute("get_offset", "getOffset");
+    layerModel.initRegularAttribute(
+      "dash_gap_pickle",
+      "dashGapPickle",
+    );
+    layerModel.initVectorizedAccessor("get_filter_value", "getFilterValue");
```
kylebarron, line 164: Oh actually you want to delete `getFilterValue`; that only applied to the data filter extension

lonboard/layer_extension.py
```diff
@@ -304,3 +304,101 @@ class DataFilterExtension(BaseExtension):
...
+    - Default False.
+
+    """
+
+    _extension_type = traitlets.Unicode("path-style").tag(sync=True)
+
+    _layer_traits = {
+        "get_dash_array": FloatAccessor(None, allow_none=True),
+        "dash_justified": traitlets.Bool(False).tag(sync=True),
+        "get_offset": FloatAccessor(None, allow_none=True),
```
kylebarron, line 380: `FloatAccessor` is correct for `get_offset`, but we need something else for `get_dash_array` because that has two values per row.

In this case we want to use an Arrow `FixedSizeList` with a width of 2, because we know that every row will have two values. For this, we want to create a new accessor trait validator. This can be mostly copied from the ColorAccessor
https://github.com/developmentseed/lonboard/blob/f597c6f19c2c7d424e02c561452543597d77e772/lonboard/traits.py#L197
You just want to change the checks for length 3 or 4  to length 2.

shriv: Added `DashArrayAccessor` in `traits.py`. Kept type checks as `int` but I'm not sure if that is correct. The `deck.gl` docs refer to a `Number` type but this can be float or int..!

src/model/extension.ts
shriv: Removed! Was an artefact from the copy!

src/model/extension.ts
shriv: Cheers. Done. Also added the import for `isDefined`. I wasn't sure what to pass in to the constructor and kept it as [`dash` only as that was in the `deck.gl` API reference](https://deck.gl/docs/api-reference/extensions/path-style-extension#constructor).

shriv:
Don't understand why one of the Python tests is failing? 🤔

kylebarron:
That's the linting on the Python code. You can run `pip install pre-commit && pre-commit run --all-files` to run that yourself.

kylebarron:
I cleaned up some of the code. I don't have the time to test this now, but it looks correct, so I'll merge and then after the next release you can test it on some of your data

Issue #513, opened 2024-05-11
Invert layers order in `split_mixed_gdf` function


RaczeQ:
What do you think about changing the order of gdfs and subsequently layers to keep points and lines on top of polygons? It seems natural to me, but I don't know if it's viable to do so.
https://github.com/developmentseed/lonboard/blob/b2ca1929c4de491c2347455c0b8717de6fc78fb1/lonboard/_utils.py#L143-L161

```python
viz(
    result_gdf.geometry, # since points are being split first, they land as first layer on the bottom
    # optional styling
    scatterplot_kwargs=dict(get_fill_color="#5c53a5"),
    path_kwargs=dict(get_color="#eb7f86"),
    polygon_kwargs=dict(stroked=False, get_fill_color="#f3e79b"),
    map_kwargs=dict(basemap_style=basemap.CartoBasemap.VoyagerNoLabels),
)
```
![image](https://github.com/developmentseed/lonboard/assets/17250607/a1fa6cc3-cf75-40fe-9151-51153e1129d7)

RaczeQ:
I've created a #516 PR for this issue 😉

kylebarron:
Closed by https://github.com/developmentseed/lonboard/pull/516

PR #516, opened 2024-05-13
Feat: reverse the layer order for automatically split geometry

RaczeQ:
## What I am changing

- Layer order for objects with mixed geometry types

## How I did it
Changed list appending order for `split_mixed_gdf` and `parse_wkb_table` function from [`points`, `linestrings`, `polygons`] to [`polygons`, `linestrings`, `points`]

## How you can test it

- Points and paths should be now plotted on top of polygon for better visibility.

Old:
Untitled

New:
image

## Related Issues
#513

### DIFF
--- a/lonboard/_geoarrow/parse_wkb.py
+++ b/lonboard/_geoarrow/parse_wkb.py
@@ -54,8 +54,8 @@ def parse_wkb_table(table: pa.Table) -> List[pa.Table]:
     if GeometryType.LINEARRING in unique_type_ids:
         raise ValueError("LinearRings not currently supported")
 
-    point_indices = np.where(
-        (type_ids == GeometryType.POINT) | (type_ids == GeometryType.MULTIPOINT)
+    polygon_indices = np.where(
+        (type_ids == GeometryType.POLYGON) | (type_ids == GeometryType.MULTIPOLYGON)
     )[0]
 
     linestring_indices = np.where(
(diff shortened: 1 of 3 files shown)


lonboard/_geoarrow/parse_wkb.py
```diff
@@ -54,28 +54,28 @@ def parse_wkb_table(table: pa.Table) -> List[pa.Table]:
     if GeometryType.LINEARRING in unique_type_ids:
         raise ValueError("LinearRings not currently supported")
 
-    point_indices = np.where(
-        (type_ids == GeometryType.POINT) | (type_ids == GeometryType.MULTIPOINT)
+    polygon_indices = np.where(
+        (type_ids == GeometryType.POLYGON) | (type_ids == GeometryType.MULTIPOLYGON)
```
kylebarron, line 58: Can you revert these lines and 66-67, since these aren't connected to the ordering?

lonboard/_geoarrow/parse_wkb.py
```diff
@@ -54,28 +54,28 @@ def parse_wkb_table(table: pa.Table) -> List[pa.Table]:
...
-    polygon_indices = np.where(
-        (type_ids == GeometryType.POLYGON) | (type_ids == GeometryType.MULTIPOLYGON)
+    point_indices = np.where(
+        (type_ids == GeometryType.POINT) | (type_ids == GeometryType.MULTIPOINT)
     )[0]
 
-    if len(point_indices) > 0:
-        point_field, point_arr = construct_geometry_array(
-            shapely_arr[point_indices],
+    if len(polygon_indices) > 0:
```
kylebarron, line 70: Before this line, can you add a comment saying something like

> We intentionally first check polygons, then line strings, then points, so that the order of generated layers is polygon, then line strings, then points. This ensures that points are rendered on top and polygons are rendered on the bottom.

In basically all my code I order my checks as point, linestring, polygon, multi point, multi linestring, multi polygon, and so I'd easily forget why we do this ordering

lonboard/_utils.py
```diff
@@ -131,21 +131,21 @@ def split_mixed_gdf(gdf: gpd.GeoDataFrame) -> List[gpd.GeoDataFrame]:
         return [gdf]
 
     if len(unique_type_ids) == 2:
-        if unique_type_ids == {GeometryType.POINT, GeometryType.MULTIPOINT}:
+        if unique_type_ids == {GeometryType.POLYGON, GeometryType.MULTIPOLYGON}:
```
kylebarron, line 134: ditto

lonboard/_utils.py
```diff
@@ -131,21 +131,21 @@ def split_mixed_gdf(gdf: gpd.GeoDataFrame) -> List[gpd.GeoDataFrame]:
...
             return [gdf]
 
-        if unique_type_ids == {GeometryType.POLYGON, GeometryType.MULTIPOLYGON}:
+        if unique_type_ids == {GeometryType.POINT, GeometryType.MULTIPOINT}:
             return [gdf]
 
     gdfs = []
-    point_indices = np.where(
-        (type_ids == GeometryType.POINT) | (type_ids == GeometryType.MULTIPOINT)
+    polygon_indices = np.where(
```
kylebarron, line 144: And ditto about the comment

kylebarron (commented):
Thanks! Just a couple nits

lonboard/_geoarrow/parse_wkb.py
RaczeQ: Sure thing, seems reasonable

lonboard/_utils.py
RaczeQ: Fixed

lonboard/_utils.py
RaczeQ: Fixed and refactored to loop as well

kylebarron (approved):
Thanks!

Issue #528, opened 2024-05-26
[BUG] Error in reprojection from GeoArrow stream with non-WGS84 crs and non-nullable coords


jorisvandenbossche:
## Context

Testing the new GeoArrow export capabilities from GeoPandas with lonboard, I ran into an issue when a reprojection happens

## Resulting behaviour, error message or logs

```python
import geopandas
from shapely import box
from lonboard import viz

df = geopandas.GeoDataFrame({"col": [1, 2]}, geometry=[box(0, 0, 1, 1), box(2, 2, 3, 3)], crs="EPSG:4326")

# those work fine
viz(df)
viz(df.to_arrow(geometry_encoding="WKB"))
viz(df.to_arrow(geometry_encoding="geoarrow"))

# with a different CRS
df2 = df.to_crs("EPSG:3857")

viz(df2)
viz(df2.to_arrow(geometry_encoding="WKB"))
# now this one fails
viz(df2.to_arrow(geometry_encoding="geoarrow"))
```

The full error:

```
---------------------------------------------------------------------------
ArrowInvalid                              Traceback (most recent call last)
Cell In[16], line 1
----> 1 viz(df2.to_arrow(geometry_encoding="geoarrow"))

File [~/conda/envs/geo-dev-gdal39/lib/python3.12/site-packages/lonboard/_viz.py:211](http://localhost:8889/home/joris/conda/envs/geo-dev-gdal39/lib/python3.12/site-packages/lonboard/_viz.py#line=210), in viz(data, scatterplot_kwargs, path_kwargs, polygon_kwargs, map_kwargs, con)
    209         layers.extend(ls)
    210 else:
--> 211     layers = create_layers_from_data_input(
    212         data,
    213         _viz_color=color_ordering[0],
    214         scatterplot_kwargs=scatterplot_kwargs,
    215         path_kwargs=path_kwargs,
    216         polygon_kwargs=polygon_kwargs,
    217         con=con,
    218     )
    220 map_kwargs = {} if not map_kwargs else map_kwargs
    222 if "basemap_style" not in map_kwargs.keys():

File [~/conda/envs/geo-dev-gdal39/lib/python3.12/site-packages/lonboard/_viz.py:303](http://localhost:8889/home/joris/conda/envs/geo-dev-gdal39/lib/python3.12/site-packages/lonboard/_viz.py#line=302), in create_layers_from_data_input(data, con, **kwargs)
    301 if hasattr(data, "__arrow_c_stream__"):
[...]

PR #532, opened 2024-05-27
Fix reprojecting GeoArrow input

kylebarron:
Closes https://github.com/developmentseed/lonboard/issues/528

When GeoPandas 1.0 is out, we should add a test for this. Ref https://github.com/developmentseed/lonboard/issues/533


kylebarron:
Published 0.9.3

PR #539, opened 2024-06-05
Document Panel in-place updates

MarcSkovMadsen:
This PR updates the Panel documentation to demonstrate that it is in fact possible to make efficient, in-place updates of a map instead of replacing a map in its entirety.

image

panel_application_example


assets/panel-application-example.gif
kylebarron, line 1: FWIW We don't usually capitalize the `B` in Lonboard

docs/ecosystem/panel.md
```diff
@@ -1,108 +1,221 @@
...
-from palettable.palette import Palette
+pn.extension("ipywidgets")
+
+@pn.cache
+def get_data():
+    return gpd.read_file(gpd.datasets.get_path("naturalearth_cities"))
+
+gdf = get_data()
+layer = ScatterplotLayer.from_geopandas(gdf, radius_min_pixels=2, get_fill_color="red")
+cities_map = Map(layer, _height=500)
```
kylebarron, line 32: Is this `_height` parameter required here? Does it work without it? Ideally the Map would automatically take up the full height of the `div` provided by Panel, but that might be tricky.

docs/ecosystem/panel.md
```diff
@@ -1,108 +1,221 @@
...
+).servable()
+```
+
+Now run:
+
+```bash
+panel serve app.py --autoreload
+```
+
+Finally, open [http://localhost:5006](http://localhost:5006). It should look like this:
```
kylebarron, line 46: Does panel always use port `5006`, or will it sometimes be on a different port?

docs/ecosystem/panel.md
```diff
@@ -1,108 +1,221 @@
...
-    ncols=3,
-    swatch_width=100,
-    name="cmap by Colorcet",
-    sizing_mode="stretch_width",
-)
-alpha = pn.widgets.FloatSlider(
-    value=0.8, start=0, end=1, name="Alpha", min_width=100, sizing_mode="stretch_width"
+
+url = "https://naciscdn.org/naturalearth/10m/cultural/ne_10m_roads_north_america.zip"
+gdf = pn.state.as_cached(
```
kylebarron, line 116: Is this the same as the decorator with `@pn.cache` in the previous example? Is one approach more or less common than the other? If they do the same things, it might be clearer to use the same syntax in both examples and refactor this line to a `get_data` function as well.

docs/ecosystem/panel.md
```diff
@@ -1,108 +1,221 @@
...
-    pn.panel(
-        pn.bind(create_map, state=state, cmap=cmap, alpha=alpha.param.value_throttled),
-        sizing_mode="stretch_both",
-    ),
-    sizing_mode="stretch_both",
-)
+
+def to_rgb(hex: str) -> list:
+    h = hex.strip("#")
+    return list(int(h[i : i + 2], 16) for i in (0, 2, 4))
```
kylebarron, line 134: Possibly a _slight_ preference for suggesting users to use [`matplotlib.colors.to_rgb`](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.to_rgb.html) instead of rolling their own hex -> RGB conversion, as we do internally [here](https://github.com/developmentseed/lonboard/blob/47464d54042e9bf7889fc6854b3b82fb1d33af5a/lonboard/traits.py#L309), but if `Palette` needs integers instead of floats then we might as well keep it as is.

kylebarron (commented):
This looks amazing! Thank you!

Ideally we would remove the `_height` parameter... I assume you tested that and it didn't work without it? Happy to merge and not block on that, but maybe with a comment somewhere saying that the `_height` parameter is subject to change in the future.

docs/ecosystem/panel.md
MarcSkovMadsen: Always port 5006 unless you change it with the `--port` flag.

docs/ecosystem/panel.md
MarcSkovMadsen: I tried and I don't get a colormap. I just get a black color.

image

assets/panel-application-example.gif
MarcSkovMadsen: Fixed

docs/ecosystem/panel.md
MarcSkovMadsen: I have removed the use of `_height` and replaced it with `map.layout.width=map.layout.height="100%"`. It will work better for both fixed and responsive sizing modes. I've done this in all 3 examples.

docs/ecosystem/panel.md
MarcSkovMadsen: I've aligned the examples to always use `@pn.cache` on a `get_data` function.

MarcSkovMadsen:
Thanks for the feedback @kylebarron. I believe I've addressed the comments. Let me know if there is more I should do. Thanks.

docs/ecosystem/panel.md
```diff
@@ -29,11 +31,14 @@ def get_data():
 
 gdf = get_data()
 layer = ScatterplotLayer.from_geopandas(gdf, radius_min_pixels=2, get_fill_color="red")
-cities_map = Map(layer, _height=500)
+cities_map = Map(layer)
+
+# Fit to the available space
+cities_map.layout.height = cities_map.layout.width = "100%"
```
kylebarron, line 37: oh cool. TIL. Maybe we should be setting this from the Lonboard side.

kylebarron (approved):

kylebarron:
Thanks! I kicked off a docs build, so this should be on the website shortly

Issue #545, opened 2024-06-13
[BUG] Layer Attribution Option


wildintellect:
## Context

What results were you expecting? <br/>
Some external layers require attribution as part of their licensing and terms of use. Many mapping libraries implement this as an `attribution` key on a layer definition that becomes a rendered over the map static text (sometimes HTML is supported).

BaseMap Layers might already have this specified in their style json, but will likely come up with many layer types particularly BitMapTileLayer where an external XYZ tile service is used.
https://maplibre.org/maplibre-style-spec/sources/#attribution
https://maplibre.org/maplibre-gl-js/docs/API/classes/AttributionControl/

kylebarron:
This is somewhat related to ideas I've had about letting users add to the underlying Maplibre map, e.g. through custom basemap styles https://github.com/developmentseed/lonboard/issues/494. But there's also a question of how tied Lonboard should be to Maplibre. It makes some things like this simpler but also makes other things harder, because tying us to maplibre ties us to Web Mercator

vgeorge:
In #561, we fixed the styles for the MapLibre attribution control and added the `custom_attribution` parameter to the Map class to allow passing a custom attribution. However, we still need to:

- Enable interactivity on the legend control (see #437)
- After that, update the docs to indicate the user should use an HTML string to add links in the attribution

prusswan:
> This is somewhat related to ideas I've had about letting users add to the underlying Maplibre map, e.g. through custom basemap styles #494. But there's also a question of how tied Lonboard should be to Maplibre. It makes some things like this simpler but also makes other things harder, because tying us to maplibre ties us to Web Mercator

Which other projections do you plan on supporting that is not web mercator? If you are considering broader support in the form of WMTS, there is OWSLib (not sure how well it will work with an interactive plot). On the JS front, OpenLayers has comprehensive WMTS support but it is probably easier to add Lonboard functionality to OpenLayers than the other way around.

kylebarron:
deck.gl can render cartesian coordinates, so in principle we can render data in _any_ coordinate system. We'll only be able to provide basemap data via Maplibre in Web Mercator, because Maplibre only supports Web Mercator. But we could have some minimal context information (country boundaries?) in other projections, to render under the data to give some context.

I've never used OpenLayers and I don't see us integrating with it.

prusswan:
[...]

Issue #436, opened 2024-03-24
Import MapLibre CSS


kylebarron:
This is required for controls and to show the AttributionControl

PR #561, opened 2024-07-01
Fix attribution

vgeorge:
This adds a `custom_attribution` property to the `Map` class and fix Maplibre CSS by adding an import. Contributes to:

- #436
- #545

To test, add the `custom_attribution` property to a Map instance as demonstrated in the API docs examples.

@kylebarron this is ready for review.

### DIFF
--- a/lonboard/_map.py
+++ b/lonboard/_map.py
@@ -176,12 +176,15 @@ class Map(BaseAnyWidget):
     """
     Custom attribution to display on the map.
 
-    This attribute supports the same format as the `attribution` property in the Maplibre API.
+    This attribute supports the same format as the `attribution` property in the
+    Maplibre API.
 
     - Type: `str` or `List[str]`
     - Default: `None`
 
-    You can provide either a single string or a list of strings for custom attributions. If an attribution value is set in the map style, it will be displayed in addition to this custom attribution.
+    You can provide either a single string or a list of strings for custom attributions.
+    If an attribution value is set in the map style, it will be displayed in addition to
+    this custom attribution.
 
     **Example:**
 
--- a/lonboard/_map.py
+++ b/lonboard/_map.py
@@ -167,6 +167,38 @@ class Map(BaseAnyWidget):
       [`lonboard.basemap.CartoBasemap.PositronNoLabels`][lonboard.basemap.CartoBasemap.PositronNoLabels]
     """
 
+    custom_attribution = traitlets.Union(
+        [traitlets.Unicode(allow_none=True), traitlets.List(traitlets.Unicode(allow_none=False))]
+    ).tag(sync=True)
+    """
+    Custom attribution to display on the map.
+
+    This attribute supports the same format as the `attribution` property in the Maplibre API.
+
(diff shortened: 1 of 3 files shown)


lonboard/_map.py
```diff
@@ -167,6 +167,44 @@ def __init__(
       [`lonboard.basemap.CartoBasemap.PositronNoLabels`][lonboard.basemap.CartoBasemap.PositronNoLabels]
     """
 
+    custom_attribution = traitlets.Union(
+        [
+            traitlets.Unicode(allow_none=True),
+            traitlets.List(traitlets.Unicode(allow_none=False)),
+        ]
+    ).tag(sync=True)
```
kylebarron, line 175: Does the attribution need to support hyperlinking too? I.e. do we need to define not just text but also the target URL?

kylebarron:
This is also quite tied to the decision here: https://github.com/developmentseed/lonboard/pull/437, of whether we should fully tie ourselves to Maplibre or not, or whether we should at least provide an option for ourselves in the future to have deck-only rendering without Maplibre

lonboard/_map.py
vgeorge: @kylebarron the MapLibre attribution control renders html links if the attribution text contains an <a> element. They are not appearing now because the Map container has `z-index: -1`, making the DeckGL canvas to be placed in front of the map controllers, blocking mouse events on it. In the recording below I disabled the `z-index`, but that caused the data layer to disappear. The z-index is not defined in Lonboard code, I would suggest we address this issue in separate because resolving the z-index conflict might require a more complex solution that ensures both the DeckGL canvas and MapLibre controllers are correctly rendered and interactive.

hyperlink

kylebarron: Right. I think this is again tied to #437. The "proper" way as of the latest deck.gl versions is to use MapboxOverlay, which has the pros and cons described in #437. I think when you use MapboxOverlay then the maplibre attribution control is clickable.

kylebarron (approved):

lonboard/_map.py
```diff
@@ -167,6 +167,44 @@ def __init__(
...
+    You can provide either a single string or a list of strings for custom attributions.
+    If an attribution value is set in the map style, it will be displayed in addition to
+    this custom attribution.
+
+    **Example:**
+
+        ```py
+        m = Map(
+            layers,
+            custom_attribution="Development Seed"
```
kylebarron, line 194: Should we maybe mention that this `custom_attribution` string can be HTML? Or should we wait on that until the link is actually clickable with fixed z-index?

lonboard/_map.py
vgeorge: Apologies, I haven't looked into #437 in depth. I believe we can implement a custom attribution control, which would untie us from MapLibre. We would need a way to extract attribution from map styles JSON, but I think it is doable. Should we keep discussing this on #561?

lonboard/_map.py
vgeorge: I believe we should mention it once it is supported properly. Let's keep tracking it on #561.

Issue #422, opened 2024-03-18
Support deck.gl v9


kylebarron:
Ref https://github.com/geoarrow/deck.gl-layers/pull/112

kylebarron:
Closed by https://github.com/developmentseed/lonboard/pull/562

PR #562, opened 2024-07-03
Upgrade Deck.gl to version 9

vgeorge:
Contributes to #422.

This is a work in progress. The `north-america-roads` example works and type checks pass, but I haven't been able to test this thoroughly yet.

cc @kylebarron @batpad


kylebarron:
Awesome! I'll take a look through some examples!

kylebarron:
Hmm it looks like the `HeatmapLayer` is broken. Which you can see from the [duckdb example notebook](https://github.com/developmentseed/lonboard/blob/main/examples/duckdb.ipynb).
image

I would initially assume this is an upstream deck.gl bug. Ah yes, looks like we're hitting this: https://github.com/visgl/deck.gl/issues/8960.

kylebarron:
9.0.23 was published with the fix for the heatmap layer so it looks like we should be able to upgrade to 9.x now!

kylebarron:
Well actually I'm getting the same error with the latest deck.gl 😔 
image

I think we need to have a better way to generate pure-JS deck.gl snapshots so that we have minimal JS projects to create deck.gl issues.

kylebarron:
> I think we need to have a better way to generate pure-JS deck.gl snapshots so that we have minimal JS projects to create deck.gl issues.

Created https://github.com/developmentseed/lonboard/issues/573

vgeorge:
@kylebarron it looks like `@geoarrow/deck.gl-layers` is using Deck.gl `v9.0.12`, which doesn't include the fix from https://github.com/visgl/deck.gl/issues/8960. Could this be the reason the heatmap layer is not loading properly?

kylebarron:
> @kylebarron it looks like `@geoarrow/deck.gl-layers` is using Deck.gl `v9.0.12`, which doesn't include the fix from [visgl/deck.gl#8960](https://github.com/visgl/deck.gl/issues/8960). Could this be the reason the heatmap layer is not loading properly?

@geoarrow/deck.gl-layers only has a peer dependency on deck.gl: https://github.com/geoarrow/deck.gl-layers/blob/25fdeafe2d9c591570bab975ebea43ae82e97a8d/package.json#L46-L49

We can verify that the deck.gl version used in this branch is indeed 9.0.23:
https://github.com/developmentseed/lonboard/blob/44146ee615ef246147adf8cd7525d4c4201a85a2/package-lock.json#L1168-L1169

kylebarron:
I added a notice in the docs and a warning in user code that the HeatmapLayer does not currently work with deck.gl v9

kylebarron:
Thanks @vgeorge for your help!

PR #567, opened 2024-07-10
Fix bbox map select feature

vgeorge:
This is based on the WIP PR https://github.com/developmentseed/lonboard/pull/417. As this introduces considerable changes, I wanted to discuss them first before adding them to the `bbox-map-select` branch.

My main goal in this PR is to fix the bbox drawing behavior. The current implementation relies on `useState` and `useMemo`, which are hard to get right due to racing conditions with the map state and do not scale well once we start working on other data management features on the map.

In this proposed approach, we use a reducer and selectors, decoupling the application logic from the component. While this introduces some boilerplate code, the Deck.gl map works more like a presentation layer, making it more maintainable and scalable in the long run.

The bbox drawing behavior seems to be working well. The workflow is similar: click on the top-right button to start selecting, then click on the map to define the start and end points.

It is possible to inspect the selected objects by opening the browser console and looking into the application state after the final click.

I haven't implemented the selected features exchange with the Python core because it is not clear to me what is the best approach. Deck.gl's [pickObjects](https://deck.gl/docs/api-reference/core/deck#pickobjects) doesn't seem to pick all objects in the bbox. It might be because it doesn't pick occluded features. The function [pickMultipleObjects](https://deck.gl/docs/api-reference/core/deck#pickmultipleobjects) seems to be the only method that picks occluded features, but the selection area must be a circle. If we find a way to get the feature IDs, we can implement the [approach](https://github.com/developmentseed/lonboard/pull/417#issuecomment-2020508957) suggested by @kylebarron, but at the moment, the only approach seems to be passing the bbox to the Python code so it can query the features.

Other known issues in this PR that we can address in the base PR:

- Map pan drag is not disabled during bbox drawing
- Cursor should update when drawing the bbox
- Reducer logger is always on; we need to figure out the best way to disable it when not in development mode

@kylebarron @batpad please let me know your thoughts on this.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,10 +1,12 @@
-import React, { useCallback, useReducer, useRef } from "react";
-import { useState, useEffect } from "react";
+import * as React from "react";
+import { useCallback, useEffect, useReducer, useRef, useState } from "react";
 import { createRender, useModelState, useModel } from "@anywidget/react";
 import type { Initialize, Render } from "@anywidget/types";
 import Map from "react-map-gl/maplibre";
 import DeckGL, { DeckGLRef } from "@deck.gl/react/typed";
-import { MapViewState, PickingInfo, type Layer } from "@deck.gl/core/typed";
+import { PolygonLayer } from "@deck.gl/layers/typed";
+import type { PickingInfo } from "@deck.gl/core/typed";
+import { MapViewState, type Layer } from "@deck.gl/core/typed";
 import { BaseLayerModel, initializeLayer } from "./model/index.js";
 import type { WidgetModel } from "@jupyter-widgets/base";
 import { initParquetWasm } from "./parquet.js";
@@ -15,9 +17,6 @@ import { Message } from "./types.js";
 import { flyTo } from "./actions/fly-to.js";
 import { useViewStateDebounced } from "./state";
 import * as selectors from "./reducer/selectors";
-
-import "maplibre-gl/dist/maplibre-gl.css";
-
 import {
   ActionTypes,
   baseInitialState,
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -196,6 +196,13 @@ function App() {
     [isMapHoverEnabled, justClicked],
   );
 
+  let buttonLabel = "Click here to start selecting";
+  if (state.mapMode === MapMode.BBOX_SELECT_START) {
+    buttonLabel = "Click the map to start drawing the selection box";
+  } else if (state.mapMode === MapMode.BBOX_SELECT_UPDATE) {
+    buttonLabel = "Click the map to finish drawing the selection box";
+  }
+
   return (
     <div id={`map-${mapId}`} style={{ height: mapHeight || "100%" }}>
       <div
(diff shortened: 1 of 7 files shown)


src/index.tsx
```diff
@@ -1,21 +1,29 @@
-import * as React from "react";
-import { useState, useEffect, useMemo, useRef } from "react";
+import React, { useCallback, useReducer, useRef } from "react";
+import { useState, useEffect } from "react";
```
kylebarron, line 2: I'd love some automated JS tool to organize imports. 

But outside of that, it would be nice to try to keep the style of the existing imports. Can we either combine these imports into one or keep the star-import and named import separate?

src/index.tsx
```diff
@@ -1,21 +1,29 @@
...
 import { getTooltip } from "./tooltip/index.js";
-import { makePolygon, isDefined, loadChildModels } from "./util.js";
+import { isDefined, loadChildModels, rateLimit } from "./util.js";
 import { v4 as uuidv4 } from "uuid";
 import { Message } from "./types.js";
 import { flyTo } from "./actions/fly-to.js";
 import { useViewStateDebounced } from "./state";
+import * as selectors from "./reducer/selectors";
+
+import "maplibre-gl/dist/maplibre-gl.css";
```
kylebarron, line 19: This is new from https://github.com/developmentseed/lonboard/commit/98ba471b2959bbd29130a67600c30975d72c3cfc right? Maybe we should merge latest main onto the target branch first?

src/index.tsx
```diff
@@ -1,21 +1,29 @@
...
+
+import {
+  ActionTypes,
+  baseInitialState,
+  MapMode,
+  reducer,
+} from "./reducer/index.js";
```
kylebarron, line 26: Move this next to previous reducer import?

kylebarron (approved):
Thinking about reducers kinda breaks my brain, but this lgtm. I'd love for @batpad to review this too

src/index.tsx
vgeorge: Apologies I missed this, I reverted to the previous styles.

src/index.tsx
vgeorge: I agree, lets merge main onto the target branch firtst. I removed the import.

src/index.tsx
vgeorge: Done.

vgeorge:
@kylebarron Thanks for the review! I agree the reducer approach can be a little hard to understand. Personally, I would prefer using [XState](https://stately.ai/docs/) for handling UI workflows. It uses state charts and makes it much easier to understand what is happening where. We can discuss introducing it if the reducer approach is too cumbersome. I'll wait for @batpad review on this before merging.

batpad:
This most definitely looks like a much better approach than whatever I had in the base PR - handling the state in the component was getting overly messy and to me, this is definitely an improvement.

I have no strong feelings on what approach we use for state management - this Reducer approach is familiar to me and seems fine, but I can also see how it can make the brain hurt a bit.

I've not tested this, but I definitely think this is good to merge into the base branch.

@vgeorge I think what would be needed before merging to `main` is having some slightly better UI elements for the bbox picking? Maybe some icons or so to start / stop selecting?

I'm 👍 to merging this into the base branch.

batpad (approved):

vgeorge:
@batpad thanks! Let's track the remaining work on the base branch.

Issue #543, opened 2024-06-12
[EPIC] Support for ColumnLayer (extruded point values)


cboettig:
## Context

I'd like to be able to use lonboard to create 3d mapping of data that include extruded column layers, [like this](https://data.source.coop/cboettig/conservation-policy/examples/ira.html) (pure pydeck code below).

![image](https://github.com/developmentseed/lonboard/assets/222586/e467ec4d-3d92-430b-8cf4-b93855325079)

## Issue

I'd like to be able to translate this pure-pydeck example into lonboard (i.e. as proof-of-principle for support for this layer).

```python
import geopandas as gpd
import pydeck as pdk

DATA_URL = 'https://data.source.coop/cboettig/conservation-policy/Inflation_Reduction_Act_Projects.geojson'
df = gpd.read_file(DATA_URL)

column_layer = pdk.Layer(
    "ColumnLayer",
    data=df,
    get_position=["LONGITUDE", "LATITUDE"],
    get_elevation="FUNDING_NUMERIC",
    get_fill_color = [256,256,0, 140],
    elevation_scale=.01,
    radius=10000,
    pickable=True,
    auto_highlight=True,
)

INITIAL_VIEW_STATE = pdk.ViewState(latitude=35, longitude=-100, zoom=4, max_zoom=16, pitch=45, bearing=0)

r = pdk.Deck(
    column_layer,
    initial_view_state = INITIAL_VIEW_STATE,
    map_style=pdk.map_styles.CARTO_ROAD,
)
r
```

I think the key piece here is just adding support for "ColumnLayer", but not sure.  Ideally I'd like to be able to layer this kind of rendering with other layers also generated in lonboard.

kylebarron:
Indeed, the primary functionality that needs to be added is support for ColumnLayer. It should be relatively straightforward.

Though on the JS side there's a question of whether to add this functionality to https://github.com/geoarrow/deck.gl-layers now that there's a semi-official fork in https://github.com/visgl/deck.gl-community that I haven't had time to engage with yet.

kylebarron:
Oh actually there's already a `ColumnLayer` in deck.gl-layers https://github.com/geoarrow/deck.gl-layers/blob/main/src/column-layer.ts. So we only need to touch the Lonboard side

kylebarron:
[...]

PR #574, opened 2024-07-16
Add `ColumnLayer`

kylebarron:
Closes https://github.com/developmentseed/lonboard/issues/543

We already had the column layer on the JS side; so this PR only adds the Python bindings + a docs page

From @cboettig 's example in the above issue:

```py
import geopandas as gpd
DATA_URL = 'https://data.source.coop/cboettig/conservation-policy/Inflation_Reduction_Act_Projects.geojson'
df = gpd.read_file(DATA_URL)

layer = ColumnLayer.from_geopandas(
    df, 
    get_elevation=df["FUNDING_NUMERIC"], 
    get_fill_color = [255, 255, 0, 140],
    elevation_scale=.01,
    radius=10000,
    pickable=True,
    auto_highlight=True,
)

m = Map(
    layer,
    view_state={"longitude": -100, "latitude": 35, "zoom": 4, "pitch": 45, "bearing": 0},
)
m
```

image
image

### DIFF
--- a/src/model/layer.ts
+++ b/src/model/layer.ts
@@ -268,9 +268,7 @@ export class ColumnModel extends BaseArrowLayerModel {
   protected radius: GeoArrowColumnLayerProps["radius"] | null;
   protected angle: GeoArrowColumnLayerProps["angle"] | null;
 
-  // // @ts-expect-error Property 'vertices' has no initializer and is not
-  // definitely assigned in the constructor
-  // Ref https://github.com/visgl/deck.gl/pull/8453
+  // Note: not yet exposed to Python
   // protected vertices: GeoArrowColumnLayerProps["vertices"] | null;
   protected offset: GeoArrowColumnLayerProps["offset"] | null;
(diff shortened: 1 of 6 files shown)


src/model/layer.ts
```diff
@@ -268,10 +268,10 @@ export class ColumnModel extends BaseArrowLayerModel {
   protected radius: GeoArrowColumnLayerProps["radius"] | null;
   protected angle: GeoArrowColumnLayerProps["angle"] | null;
 
-  // @ts-expect-error Property 'vertices' has no initializer and is not
+  // // @ts-expect-error Property 'vertices' has no initializer and is not
```
vgeorge, line 271: @kylebarron should we clean up these commented lines?

kylebarron: I'll add a comment here that the commented out lines represent deck.gl properties that we don't currently expose to Python

kylebarron:
I'll add a quick example tomorrow and then merge

kylebarron:
I'm going to merge before I do a potential refactor, and add a note to include an example before the next release

Issue #581, opened 2024-07-24
Remove pyarrow as hard dependency


kylebarron:
**Is your feature request related to a problem? Please describe.**

pyarrow is a massive, monolithic dependency. It can be hard to install in some places, and can't currently be installed in Pyodide. It's certainly a monumental effort to get it to work in Pyodide, but I think it would be valuable for lonboard to wean off of pyarrow.

The core enabling factor here is the [Arrow PyCapsule Interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html). It allows Python Arrow libraries to exchange Arrow data _at the C level_ at no cost. This means that we can interface  at no cost with any user who's already using pyarrow, but not be required to use pyarrow ourselves. I've been promoting its use throughout the Python Arrow ecosystem (https://github.com/apache/arrow/issues/39195#issuecomment-2245718008), and hoping this grows into something as core to tabular data processing as the buffer protocol is to numpy.

As part of working to build the ecosystem, I created [arro3](https://github.com/kylebarron/arro3), a new, very minimal Python Arrow implementation that wraps the Rust Arrow implementation.

I think that it should be possible to swap out pyarrow for arro3, which is about 1% of the normal pyarrow installation size.

It's also symbiotic for the ecosystem if Lonboard shows the benefits of modular Arrow libraries in Python.

**Describe the solution you'd like**

We'll keep pyarrow as a required dependency for GeoPandas/Pandas interop. pyarrow has implemented `pyarrow.Table.from_pandas` and that's not something I want to even think about replicating.

But aside from that, pretty much everything is doable in arro3 and geoarrow-rust.

- [ ] [`pa.Table.from_arrays`](https://github.com/developmentseed/lonboard/blob/7eb19e09e0fd83b7f71b6cc745e1ea42bb410083/lonboard/_cli.py#L56)
- [ ] [Construct a table from named columns](https://github.com/developmentseed/lonboard/blob/12300ad5f1bbea6e0d696e6bcf68403ca7e5186b/lonboard/_serialization.py#L54)
- [ ] [Write Parquet with specified compression and compression level](https://github.com/developmentseed/lonboard/blob/12300ad5f1bbea6e0d696e6bcf68403ca7e5186b/lonboard/_serialization.py#L33-L38)
[...]

PR #582, opened 2024-07-25
Remove pyarrow dependency

kylebarron:
For #581


kylebarron:
After a lot of work here and in arro3, we're pretty close to having this working! I think just three tasks!

- [ ] table rechunking so that each Parquet buffer can have reasonably similar sizes https://github.com/developmentseed/lonboard/blob/43701868c699d2fab422f1c48ca045df3d323f5b/lonboard/_serialization.py#L34
- [ ] Support for adding an array or a chunked array in `table.add_column` https://github.com/developmentseed/lonboard/blob/a60adbcbde56d665e29ac672b3918960b8d49d86/lonboard/_viz.py#L453 
- [ ] inferring the data type for primitive data types in `from_numpy` https://github.com/developmentseed/lonboard/blob/ac5baac5be8ddadd10a6c669d0fafbcbdf6b08d6/lonboard/traits.py#L1008-L1013

kylebarron:
This is now working! 
image

kylebarron:
All the tests now pass (locally)! arro3 isn't yet published to pypi, so that's why CI here is failing

kylebarron:
I ran through a couple examples manually and it's still working as expected!

PR #601, opened 2024-08-19
Use arro3 for dictionary encoding in apply_categorical_cmap

kylebarron:


kylebarron:
With https://github.com/kylebarron/arro3/pull/159 I should be able to update and merge this

PR #610, opened 2024-08-22
Add quak integration

kylebarron:
Learnings so far:

- deck.gl crashed when I supplied a column that was either 0 or 1, but it seems to work fine when it's either 1 or 2. This function https://github.com/visgl/deck.gl/blob/fdbb88a0c0cd662e95f5b6a84a36ec13ab9cc5d9/modules/extensions/src/data-filter/data-filter-extension.ts#L393-L408 is erroring with an invalid attempt to access `0`.
- I need to make sure the extension is correctly added to the layer, and that the extension's added traits are correctly initialized.
- It would be ideal if we could debounce this a bit. I suppose we could have a global variable of "last quak update", and then at the start of this `observe` callback, we can check if the time is within a certain amount of time since the last update? Not sure how that would work.

### DIFF
--- a/examples/internet-speeds.ipynb
+++ b/examples/internet-speeds.ipynb
@@ -30,15 +30,29 @@
    "outputs": [],
    "source": [
     "from pathlib import Path\n",
+    "import pyarrow as pa\n",
     "\n",
+    "from arro3.core import Array, Table, DataType, ChunkedArray\n",
     "import geopandas as gpd\n",
     "import numpy as np\n",
     "import pandas as pd\n",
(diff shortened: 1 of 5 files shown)


kylebarron:
Should look here for reference: https://discord.com/channels/1178781616714297576/1179097041134166096/1276969495151251506

kylebarron:
Note to self: you don't need the category data filter for rendering booleans. You can use the existing data filter extension and just set the range appropriately. Note that in theory this will expand the bits to float32, right?

Issue #592, opened 2024-08-12
Merging additions in `ecoscope` fork


atmorling:
Things I've added in our fork that may be worth adding/reviewing here

#### Things that are simple(ish) to merge
- v9 bump 
- Add support for Deck widgets
- Add `mapWidth` to `Map`
- An alternative `viz` that returns a layer rather than a map
- Surfacing deck.gl `MapView` in `index.tsx` - we're doing this to allow `repeat=true`
- Surface `controller` Python side in `Map`

#### Less simple
- Supporting custom widgets (non deck.gl native js components). It's easy enough to add custom components on the python side, and it's easy enough to split our custom js components into an NPM package. What's unclear to me is the best way to tie a custom js package into `lonboard` and if that's even a good idea.

batpad:
+cc @vgeorge for visibility

kylebarron:
See #613 in which @vgeorge started a PR to add your additions to lonboard.

> * An alternative `viz` that returns a layer rather than a map

In terms of API design, what do you think about just documenting that you can access the generated layers via

```py
from lonboard import viz

viz_layers = viz(data).layers
```

Then we don't need another top-level function just to create layers.

> * Surface `controller` Python side in `Map`

It looks like you made this boolean? So you didn't have a use case of customizing the controller; just of turning off interaction?

> * Add support for Deck widgets

I see that all of your deck widgets subclass from `lonboard.base.BaseWidget`. For some of the widgets that don't need _jupyter-based interactivity_, I think it might be better if they aren't their own Jupyter widgets, but rather you just pass in props to the top-level Map.

I have to think about this a little more.

atmorling:
> In terms of API design, what do you think about just documenting that you can access the generated layers via
> 
> ```python
> from lonboard import viz
> 
> viz_layers = viz(data).layers
> ```
In a world where I already have a map with some layers, and I want to add another layer based on the output of `viz` (ie I have a mixed geometry GDF, but specific styling needs for each geometry type), it would be nice to not have to create another map entirely. I guess it depends if you see that being a typical use case or not. 


> It looks like you made this boolean? So you didn't have a use case of customizing the controller; just of turning off interaction?
> 
Yep

> > * Add support for Deck widgets
> 
> I see that all of your deck widgets subclass from `lonboard.base.BaseWidget`. For some of the widgets that don't need _jupyter-based interactivity_, I think it might be better if they aren't their own Jupyter widgets, but rather you just pass in props to the top-level Map.
[...]

PR #613, opened 2024-08-27
Ecoscope fork merge

vgeorge:
This PR addresses issue #592. The aim is to integrate changes from the [ecoscope fork](https://github.com/wildlife-dynamics/lonboard) of Lonboard and discuss their potential inclusion in the main project. 

The fork was initiated in April 2024. While the front-end has evolved since then, the changes appear manageable for integration. Current commit resolves merge conflicts and includes a minor adjustment to ensure map component rendering.

Below is a summary of changes from the remote repository. Each item requires review before consideration for merging:

- [x] v9 bump
- [ ] Add mapWidth to Map
- [ ] Add support for Deck widgets
- [ ] Implement an alternative visualization that returns a layer instead of a map
- [ ] Surface deck.gl MapView in index.tsx (to enable repeat=true)
- [ ] Expose controller on the Python side in Map

@kylebarron @batpad please let me know if you have comments.


lonboard/static/index.js
kylebarron, line 1: This is generated code and should not be checked in.

lonboard/static/index.css
kylebarron, line 1: Ditto, this is also generated

package.json
```diff
@@ -17,6 +17,8 @@
     "framer-motion": "^11.3.30",
     "lodash.debounce": "^4.0.8",
     "lodash.throttle": "^4.1.1",
+    "@deck.gl/widgets": "^9.0.16",
```
kylebarron, line 20: Nit: can we alphabetically sort our dependencies? I'd like to visually keep track of deck.gl dependencies next to each other.

Also, all deck.gl libraries should have the same version constraint.

src/model/deck-widget-models.ts
```diff
@@ -0,0 +1,327 @@
+import type { WidgetModel } from "@jupyter-widgets/base";
+import { BaseModel } from "./base.js";
+import { Widget, WidgetPlacement } from "@deck.gl/core";
+import { CompassWidget, ZoomWidget, FullscreenWidget, LightTheme } from "@deck.gl/widgets";
+import { LegendWidget, NorthArrowWidget, TitleWidget, ScaleWidget, SaveImageWidget } from "./deck-widget.js";
+
+export abstract class BaseDeckWidgetModel extends BaseModel {
```
kylebarron, line 7: At first glance, I wouldn't think these should be models... do the widgets themselves need to be reactive on the Jupyter side?

kylebarron:
This is stale and we should probably start from scratch if we wanted to try and unite these forks still.

A complicating factor is that I rewrote history in lonboard to remove large files that had accidentally been added #820 so now there's extra divergence

Issue #586, opened 2024-07-30
Tooltip viewer can be too crowded


kylebarron:
When you click on an object, maybe show a sidebar or a collapsible menu.

Overture is a good example. We might want a JSON viewer

vgeorge:
In #636, we replaced the tooltip with a scrollable side panel. Currently, there is no visual indication that a feature can be clicked to open the side panel, which was not necessary when the tooltip was used.

@kylebarron you suggested documenting this behavior. Do you have any recommendations on the best approach? I still believe it is possible to change the mouse pointer on feature hover to indicate interactivity. We can do something using `getCursor` like in [here](https://github.com/developmentseed/lonboard/blob/44a45f87e7f3d7b3656966b0b0c2b0089000fe4f/src/index.tsx#L235), but it might be better to track as an enhancement in a separated ticket.

kylebarron:
I think documenting it in the release notes and documentation website is fine and sufficient

ymoisan:
The side panel is a nice addition to lonboard.  However, I believe there is still a need for a tooltip, albeit with selectable contents as opposed to dumping it all in the side panel.  I have included a video showing the behaviour I'm about to make the following comments on.

- I play with `_height` to show the effect of toggling expand/collapse cell output of the map; I'll note that with a height greater than the default cell output height, the toggle has to be "convinced" a little (like click three times before I can see the full map height), but that's not a big deal
- I like the `auto_highlight` feature a lot;; when the side panel lets the user see enough of the image to notice the current panel is linked to the highlighted feature, all is fine; however
- when the side panel covers the entire map, we have a problem; the 'x' becomes a kind of '<' because it is hidden by the bbox button in the top right

Playing with the notebook panel width (e.g. showing the lefthand side Jupyterlab panel) shows how the panel wraps content well.  So I'm thinking : is there a way to restrict the side panel to a percentage of the map cell output so as to still be able to see the selected/highlighted feature ?

As to the tooltip functionality, I believe it would still be useful for quick identification of features, when one does not want the whole pedigree.  `show_tooltip` set to True does not show a tooltip.  I think if one were to define a tooltip object attached to the map (or layer ?) it should be selectable on the map like the [PMTiles Viewer](https://pmtiles.io/) Popup/Show Attributes.  I'll try to have a look but you guys are probably in a better position to do an eventual PR.

Thanx again for lonboard, really.

[...]

PR #636, opened 2024-09-18
Replace tooltip with side panel

vgeorge:
This addresses #586 by adding a scrollbar to the tooltip when there is a long list of properties. Work in progress.

We discussed adding a side panel for browsing the properties, but I wanted to explore an alternative that enhances the existing tooltip rather than introducing a new UI element.

In this implementation, the `getTooltip` property in Deck.gl has been replaced with a custom tooltip rendered separately. This change was necessary because the trigger for displaying the tooltip was modified from mouse hover to mouse click. As a result, the tooltip persists when the mouse pointer moves away, allowing the user to scroll through the available properties.

### Known Issues:

- This is based off release `0.10.0-beta.2` commit, the main branch appears to be broken due to dependencies updates
- A horizontal scrollbar is always displayed alongside the vertical one due to an issue determining the table width, this needs an CSS update
- The tooltip remains in the same position when the map is panned. We could either hide it or have it move along with the map
- We are moving away from a Deck.gl-based solution. It might be worth implementing this as a custom widget, but the complexity of doing so is unclear at this point

### How to test

Open examples `north-america-roads` and `internet-speeds` and click on a rendered feature. The popup should be displayed and the table should be scrollable when the list is long.

cc @kylebarron @batpad


kylebarron:
Looks fine at a glance in principle.

I'd suggest including screenshots or screencasts in PRs. It makes it easier to consider the changes without pulling this branch, rebuilding, bringing up an example notebook, and running it. It also preserves the semantic meaning of this PR for us to look back at in the future.

> * This is based off release `0.10.0-beta.2` commit, the main branch appears to be broken due to dependencies updates

Maybe that's from https://github.com/developmentseed/lonboard/pull/629. Would you be able to try downgrading and see if it fixes your issue?

> * The tooltip remains in the same position when the map is panned. We could either hide it or have it move along with the map

Remains in the same pixel position or in the same geospatial position?

vgeorge:
@kylebarron and I had a chat about this. Some notes:

- Having a side panel would be better for inspecting the data, as it won't cover nearby features like the tooltip does. We'll  shift from a tooltip to a scrollable sidebar
- With the removal of the tooltip there would be no indication of the highlighted feature and Kyle suggested looking into the [autoHighlight Layer property](https://deck.gl/docs/api-reference/core/layer#autohighlight)

cc @batpad

vgeorge:
## Changes Added

- Merged the latest commit from the main branch, fixing the dependency issue
- Replaced the tooltip with a side panel

## Current Behavior

- Clicking on a feature on the map opens a left-side panel, adjusting the map width accordingly
- A button is available to close the panel
- A scrollbar is displayed if necessary
- BBox selection works regardless of whether the panel is open or closed

### Internet Speeds Example

internet-speeds

### North American Roads Example

roads

## Feature Highlight

After setting `auto_highlight=True`, I observed the following behavior:

autohighlight

While light features are distinguishable, darker ones are not, so I’m unsure if this is the best approach.

We can explore changing the mouse pointer style to `pointer` to better indicate that it is possible to click on a feature. However, this is not as straightforward as simply enabling auto highlight, so we might need to address this separately if we want to include the side panel feature in the next release.

@kylebarron please let me know your toughts.

kylebarron:
This looks amazing!!

> After setting `auto_highlight=True`, I observed the following behavior:
>
> While light features are distinguishable, darker ones are not, so I’m unsure if this is the best approach.

I don't think this is a problem because the end user can set a custom `highlightColor`: https://deck.gl/docs/api-reference/core/layer#highlightcolor

(Well, it looks like we need to actually expose that through Lonboard)

> We can explore changing the mouse pointer style to `pointer` to better indicate that it is possible to click on a feature.

In terms of optimal UI, it seems like it would be great to switch between "normal cursor" and "pointer cursor" depending on whether there's something active under the cursor to click on. However since deck.gl is in a WebGL canvas, I'm almost certain that's not possible.

> However, this is not as straightforward as simply enabling auto highlight, so we might need to address this separately if we want to include the side panel feature in the next release.

I think it's fine to just document new behavior that you need to click instead of just highlight on objects. I don't want to enable auto highlight by default because it doesn't work properly on multi-geometries.

kylebarron (approved):

vgeorge:
@kylebarron thanks for the review! I updated the PR title to better describe what has changed. I'll repost your comments on the upstream ticket to follow up the discussion.

Issue #675, opened 2024-10-07
Avoid sidebar resize when clicking points


kylebarron:
Right now there's some jitter when clicking on geometries because the sidebar resizes based on the size of the columns:

https://github.com/user-attachments/assets/03155387-ff27-4d6b-ae2d-7eb5c44679b8

@vgeorge any thoughts?

kylebarron:
I'm also surprised that there's full half-second lag to update the sidebar. It looks like every time you click there's a full map re-render (there's a flash), which is very bad.

This is possibly linked to the sidebar resizing if it forces the size of the deck.gl canvas to reset.

vgeorge:
@kylebarron The component re-renders on any view size change, not only when the sidebar resizes. You can confirm this by resizing the browser window without clicking any element. Looking at a [Deck.gl example](https://deck.gl/examples/arc-layer), it doesn’t re-render on window resize, so the issue is very likely in our component implementation and probably existed before the sidebar.

![resize](https://github.com/user-attachments/assets/df65efcc-c89d-401a-a931-6ea7c9243eed)


We can mitigate it by setting a fixed width to the sidebar, but it will still flash when closed and opened. The delay in showing the sidebar might be related to this.

kylebarron:
@vgeorge is this closed by #701?

vgeorge:
@kylebarron I think so. The sidebar jittering should be gone, although the re-render issue is not fixed by #701 because it is not related to the sidebar implementation.

PR #701, opened 2024-11-07
Improve side panel layout

vgeorge:
Contributes to #586 and  #675. Currently, the side panel has a variable width that changes according to the length of the property values, causing layout shifts, occupying too much screen space, and generally providing a suboptimal UI experience.

This PR aims to improve that by:

- Setting a fixed width of 24rem
- Adding a small space (4px) between the tooltip and the map
- Truncating property names
- Breaking words/lines in property values
- Improving scroll behavior

Demo, non-scrolled:

non-scrolled

Scrolled:

scrolled

@kylebarron can you please review?


kylebarron (approved):
This looks great! Thanks!

Issue #136, opened 2023-10-18
Integrate `mapclassify`


kylebarron:
From the [geopandas.plot docstring](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.plot.html#geopandas.GeoDataFrame.plot)

> Name of a choropleth classification scheme (requires mapclassify). A mapclassify.MapClassifier object will be used under the hood. Supported are all schemes provided by mapclassify (e.g. ‘BoxPlot’, ‘EqualInterval’, ‘FisherJenks’, ‘FisherJenksSampled’, ‘HeadTailBreaks’, ‘JenksCaspall’, ‘JenksCaspallForced’, ‘JenksCaspallSampled’, ‘MaxP’, ‘MaximumBreaks’, ‘NaturalBreaks’, ‘Quantiles’, ‘Percentiles’, ‘StdMean’, ‘UserDefined’). Arguments can be passed in classification_kwds.

davidbailey:
The following worked for me:

```layer = SolidPolygonLayer.from_geopandas(zones_interpolated, opacity=0.2)
jc = JenksCaspall(
    zones_interpolated["ProductionAttractionSum Density"].fillna(0),
    k=20
)
colors = []
for color in jc.yb:
    colors.append(Viridis_20.colors[color])
layer.get_fill_color = numpy.uint8(colors)

Map(layers=[layer])```

**Comment by kylebarron** [comment]
Thanks! By the way, I'm not sure where `Viridis_20` comes from, but doing an array lookup would be much faster than a loop. You should probably be able to do something like

```py
Viridis_20.colors[jc.yb]
```

knaaptime:
since this is still open, i've cooked up a proof of concept that uses pandas's extension API to create a geopandas `explore` method that uses lonboard as a backend. The code is [here](https://github.com/oturns/geosnap/pull/412/files) and an example notebook [here](https://gist.github.com/knaaptime/6f6c4f1ae7ce6b156b4cfbcc6b029f98). The functions are fairly small and dont do much besides handle classification schemes and categorical variables using mapclassify/matplotlib and expose a few more styling args for convenience and parity with the existing explore

works well for my use cases and I'm happy to keep it in geosnap if it's out of scope. That said, I suspect it might be more useful/visible in a central location like lonboard itself so also more than happy to contribute it if anyone has interest

knaaptime:
[...]

PR #728, opened 2025-01-07
feat: geopandas extension

knaaptime:
to begin review. The notebook at examples/explore should run through and demonstrate functionality. I need to take a look at how you prefer to test against Maps


knaaptime:
resolves #136

pyproject.toml
```diff
@@ -44,6 +44,7 @@ classifiers = [
 cli = ["click>=8.1.7", "pyogrio>=0.8", "shapely>=2"]
 geopandas = ["geopandas>=0.13", "pandas>=2", "shapely>=2"]
 movingpandas = ["movingpandas>=0.17"]
+explore = ["mapclassify>=2.8.1"]
```
kylebarron, line 47: I think we can include this in the `geopandas` extra

knaaptime: ok, wasnt sure how you wanted to handle that part. Geopandas itself only has a soft dependency on mapclassify so its only imported when `scheme` is used; wasn't sure if you wanted to keep it similarly isolated (even though I hadn't structured the import like that)

lonboard/geopandas.py
```diff
@@ -0,0 +1,324 @@
...
+        alpha=1,
+        layer_kwargs=None,
+        map_kwargs=None,
+        classify_kwargs=None,
+        nan_color=[255, 255, 255, 255],
+        color=None,
+        wireframe=False,
+        tiles="CartoDB Darkmatter",
+        m=None,
+    ):
```
kylebarron, line 42: Can you expand on what differences (if any) there are with the upstream geopandas.explore method? I think we'd want it to be as similar as possible.

That would be great to include in this docstring.

Additionally, ideally there would be a way to get valid type hinting here, but presumably that won't work if this is added at runtime to the GeoDataFrame instance.

lonboard/geopandas.py
```diff
@@ -0,0 +1,324 @@
...
+        elif scheme is None:
+            # minmax scale the column first, matplotlib needs 0-1
+            transformed = (gdf[column] - np.nanmin(gdf[column])) / (
+                np.nanmax(gdf[column]) - np.nanmin(gdf[column])
+            )
+            color_array = apply_continuous_cmap(
+                values=transformed, cmap=colormaps[cmap], alpha=alpha
+            )
+        else:
+            color_array = get_color_array(
```
kylebarron, line 276: This is the only place where `get_color_array` is used, right? We should only import it within this block.

lonboard/geopandas.py
knaaptime: sure happy to add both. In general I tried to keep all possible arguments the same (hence the `tiles` that uses a str --> class--> str, which isn't ideal but I'm hard conditioned to type tiles='CartoDB Positron') any time I explore.

the one thing I know is different is `classification_kwargs` used here vs `classification_kwds` used in geopandas, but I'll switch it to martin's version for compat. Otherwise, any arguments available in explore have the same name. Those that *don't* map between the two are unique to the backengs, e.g. `layer_kwargs` here vs `style_kwds` in vanilla explore

also I generally use numpy docstrings but I think the rest of the packages uses another format, so I should probably update that too

pyproject.toml
kylebarron: Hmm, in that case, maybe it is better to keep it separate. Especially if we move the `mapclassify` import only within the `else` clause where it's used. 

In that case, it's probably fine to not even define `mapclassify` in an extra.

martinfleis:
Keep me in the loop. Happy to do a proper review of this from the geopandas perspective once @kylebarron is happy.

lonboard/geopandas.py
```diff
@@ -273,14 +272,23 @@ def _dexplore(
                 values=transformed, cmap=colormaps[cmap], alpha=alpha
             )
         else:
+            try:
+                from mapclassify.util import get_color_array
+            except ImportError as e:
+                raise ImportError(
+                    "you must have the `mapclassify` package installed to use this function"
```
kylebarron, line 279: Maybe more clear?

```
"you must have the `mapclassify` package installed if you pass `classification_kwds`"
```

pyproject.toml
```diff
@@ -42,9 +42,8 @@ classifiers = [
 
 [project.optional-dependencies]
 cli = ["click>=8.1.7", "pyogrio>=0.8", "shapely>=2"]
-geopandas = ["geopandas>=0.13", "pandas>=2", "shapely>=2"]
+geopandas = ["geopandas>=0.13", "mapclassify>=2.8.1", "pandas>=2", "shapely>=2"]
```
kylebarron, line 45: I think we can take out the dependency altogether from here. If someone hits the error they'll be instructed to install it.

lonboard/geopandas.py
```diff
@@ -0,0 +1,353 @@
...
+    map_kwargs=None,
+    classification_kwds=None,
+    nan_color=[255, 255, 255, 255],
+    color=None,
+    wireframe=False,
+    tiles="CartoDB Darkmatter",
+    highlight=False,
+    m=None,
+):
+    """explore a dataframe using lonboard and deckgl
```
kylebarron, line 150: If this docstring is exactly the same as the public-facing docstring, you can just write here to refer to the public docstring for parameter information.

lonboard/geopandas.py
```diff
@@ -0,0 +1,353 @@
...
+        nan_color=[255, 255, 255, 255],
+        color=None,
+        wireframe=False,
+        tiles="CartoDB Darkmatter",
+        highlight=False,
+        m=None,
+    ):
+        """explore a dataframe using lonboard and deckgl
+
+        Parameters
```
kylebarron, line 45: Yeah we do use the other docstring format in lonboard (I can't remember which of numpy/google format is which), and it would be good to standardize on that format.

knaaptime:
i think this should be ready, though I'm not the best with type hints or google docstrings

pyproject.toml
kylebarron: I removed this extra in favor of producing a runtime error

kylebarron:
Can you check why pre-commit is still failing? You should be able to run it locally with
```
pip install pre-commit
pre-commit run --all-files
```

kylebarron:
@martinfleis do you have any comments on this PR?

I'd like to get in #753, which in general raises the code quality of the code base, and then ensure that this PR meets that bar as well.

lonboard/geopandas.py
```diff
@@ -0,0 +1,296 @@
...
+        map_kwargs: Optional[Dict[str, Any]] = None,
+        classification_kwds: Optional[Dict[str, Any]] = None,
+        nan_color: Optional[Union[List[int], np.ndarray[int]]] = [255, 255, 255, 255],
+        color: Optional[str] = None,
+        vmin: Optional[float] = None,
+        vmax: Optional[float] = None,
+        wireframe: bool = False,
+        tiles: Optional[str] = None,
+        highlight: bool = False,
+        m: Optional[Map] = None,
```
martinfleis, line 45: A personal preference here would be to move these before any of the kwargs dicts.

lonboard/geopandas.py
```diff
@@ -0,0 +1,296 @@
...
+    See the public docstring for detailed parameter information
+
+    Returns
+    -------
+    lonboard.Map
+        a lonboard map with geodataframe included as a Layer object.
+
+    """
+
+    providers = {
```
martinfleis, line 147: A minor note on inconsistency that may bite some users - geopandas is able to consume the strings representing tile names in any formatting xyzservices is capable of parsing. So both `cartodbpositron` and `CartoDB Positron` result in the same tiles.

lonboard/geopandas.py
```diff
@@ -0,0 +1,296 @@
...
+        Keyword Args:
+            column : Name of column on dataframe to visualize on map.
+            cmap : Name of matplotlib colormap to use.
+            scheme : Name of a classification scheme defined by mapclassify.Classifier.
+            k : Number of classes to generate. Defaults to 6.
+            categorical : Whether the data should be treated as categorical or
+                continuous.
+            elevation : Name of column on the dataframe used to extrude each geometry or
+                an array-like in the same order as observations. Defaults to None.
+            extruded : Whether to extrude geometries using the z-dimension.
```
martinfleis, line 58: I probably don't follow here. Shouldn't extruded be automatically True when elevation is provided? Like the two are tied together, no? Or can I use extruded without elevation? What does it extrude on, then, Z coordinate? How?

martinfleis (commented):
Looks great! Just a few minor notes.

lonboard/geopandas.py
kylebarron: Well, that is a good point that the docs below refer to these as keyword args, but in actuality all of the above parameters are positional arguments (there's no `*` in the signature). I'd be inclined to match the geopandas signature as much as possible, but then make any _added_ parameters be keyword-only.

lonboard/geopandas.py
```diff
@@ -0,0 +1,296 @@
...
+            wireframe,
+            tiles,
+            highlight,
+            m,
+        )
+
+
+def _dexplore(
+    gdf,
+    column=None,
```
kylebarron, line 115: For the internal function, we should make all of these keyword-only arguments to ensure we're passing the correct names below.
```suggestion
    *,
    column=None,
```

lonboard/geopandas.py
kylebarron: > xyzservices is capable of parsing

Do you know how that parsing works? Does it remove any whitespace and then convert to lower case?

lonboard/geopandas.py
knaaptime: i agree this is a little awkward. There's probably no need for the bool, though I was just trying to follow the viz/Layer API which treats them separately. Happy to go with consensus here.

Another small wrinkle is that sometimes in GIS or CAD software (IIRC anyway) 'elevation' sometimes refers to the height *beneath* the polygon, e.g. so you can drop the extruded poly onto a DEM for things like viewsheds. In that case I want to say this parameter is called 'extrusion_height' or something? No real opinion here, just something to consider in case folks are coming from that world

lonboard/geopandas.py
knaaptime: interesting, i had no idea those things work 😆

kylebarron:
I merged #753; would you be able to merge in latest main?

knaaptime:
looks like the only remaining failures are from the demo notebook. Could probably fix many of them, but seems a little overkill to add type annotations to some quick functions in an example notebook

knaaptime:
fixed annotations in the notebook so im not sure what the lingering failure is

lonboard/geopandas.py
knaaptime: i removed the bool so that its turned on automatically if elevation is passed

knaaptime:
copiolt thinks the failure is a caching issue (apparently), so i'd guess it will pass if re-run on its own

knaaptime:
looks like maybe some other parts of the stack are moving at the moment, but let me know if there's anything I can do to help move this forward :)

lonboard/geopandas.py
```diff
@@ -0,0 +1,340 @@
...
+    @staticmethod
+    def _validate(obj) -> None:  # noqa: ANN001
+        if not isinstance(obj, gpd.GeoDataFrame):
+            raise TypeError("must be a geodataframe")
+
+    def explore(  # noqa: PLR0913
+        self,
+        column: str | None = None,
+        cmap: str | None = None,
+        scheme: str | None = None,
```
kylebarron: Can we make some of these parameters keyword-only? Perhaps change to
```suggestion
    def explore(  # noqa: PLR0913
        self,
        column: str | None = None,
        *,
        cmap: str | None = None,
        scheme: str | None = None,
```
so that everything after `column` can only be passed as a named kwarg?

lonboard/geopandas.py
```diff
@@ -0,0 +1,340 @@
...
+    classification_kwds,  # noqa: ANN001
+    nan_color,  # noqa: ANN001
+    color,  # noqa: ANN001
+    vmin,  # noqa: ANN001
+    vmax,  # noqa: ANN001
+    wireframe,  # noqa: ANN001
+    tiles,  # noqa: ANN001
+    highlight,  # noqa: ANN001
+    m,  # noqa: ANN001
+) -> Map:
```
kylebarron: It's not ok to ignore all of these errors. At a minimum we need every function parameter and return type to be correctly typed so that we can catch internal type errors.

kylebarron: Instead of defining this as a separate function; put this implementation inside of `LonboardAccessor.explore`. There's not really any reason to have this defined as a standalone function?

lonboard/geopandas.py
```diff
@@ -0,0 +1,323 @@
...
+        path_kwargs=layer_kwargs,
+        map_kwargs=map_kwargs,
+    )
+    if m is not None:
+        new_m = m.add_layer(new_m)
+
+    return new_m
+
+
+def _get_categorical_cmap(categories, cmap, nan_color, alpha):  # noqa: ANN001, ANN202
```
kylebarron: This needs to be typed.

lonboard/geopandas.py
```diff
@@ -0,0 +1,323 @@
...
+    try:
+        from matplotlib import colormaps
+    except ImportError as e:
+        raise ImportError(
+            "this function requires the `lonboard` package to be installed",
```
kylebarron:
```suggestion
            "this function requires the `matplotlib` package to be installed",
```

lonboard/geopandas.py
```diff
@@ -0,0 +1,323 @@
...
+    *,
+    column,  # noqa: ANN001
+    cmap,  # noqa: ANN001
+    scheme,  # noqa: ANN001
+    k,  # noqa: ANN001
+    categorical,  # noqa: ANN001
+    elevation,  # noqa: ANN001
+    elevation_scale,  # noqa: ANN001
+    alpha,  # noqa: ANN001
+    layer_kwargs,  # noqa: ANN001
```
kylebarron: It looks like `layer_kwargs` is something you're adding that doesn't exist in the upstream `GeoDataFrame.explore`. In that case, you should follow the same behavior as in `viz` and use the same parameters:

https://github.com/developmentseed/lonboard/blob/c6c320957511e87ea29db8c65e08c4de5230ec16/lonboard/_viz.py#L92-L95

This will also fix your type hinting below:
image

kylebarron (changes requested):
Sorry that this has sat for a while without a review. I have too many projects and haven't dedicated time to Lonboard recently.

Because I have too many projects, I also have a high bar for contributions so that they don't grow to become a larger maintenance burden in the future.

There's still a decent amount of work here before it's review ready. In particular, there should be virtually no `type: ignore` statements.

knaaptime:
no sweat! thanks for the comments

right there with you and similarly spread thin :). Static typing is pretty extraneous in most of my projects so not very practiced; will take another pass, or several

knaaptime:
@kylebarron I think this is ready for another review, though it looks like i also pulled in a bunch of other stuff by accident. Maybe i should close this and open a new PR that has only the new file (and notebook)?

knaaptime:
> In particular, there should be virtually no `type: ignore` statements.

Everything is typed now, so the only linter complaints are the four on line 55, all of which are complexity-related and im not sure there's a reasonable way to satisfy those otherwise

kylebarron:
> @kylebarron I think this is ready for another review, though it looks like i also pulled in a bunch of other stuff by accident. 

Actually that was my fault. In fixing https://github.com/developmentseed/lonboard/issues/820, which removed 144MB of files from the git repo, I had to rewrite history. Sorry about that, it was an accident to commit those files originally.

> Maybe i should close this and open a new PR that has only the new file (and notebook)?

That would be ideal if you could do that. Hopefully it's not too much work? 😬

PR #731, opened 2025-01-14
use dependency state in export

knaaptime:
resolves #726 again


kylebarron:
It looks like this is documented [here](https://ipywidgets.readthedocs.io/en/8.0.7/embedding.html#python-interface):

> In all embedding functions in `ipywidgets.embed`, the state of all widgets known to the widget manager is included by default. You can alternatively pass a reduced state to use instead. This can be particularly relevant if you have many independent widgets with a large state, but only want to include the relevant ones in your export. To include only the state of specific views and their dependencies, use the function `dependency_state`

kylebarron:
Thanks!

PR #736, opened 2025-01-22
feat: Reinstate tooltip

vgeorge:
Contributes to #586. 

@kylebarron this just restore the original tooltip files, re-enabling the feature that was removed in #636. I think the tooltip need improvements on the layout to break lines in a similar way the sidepanel does. But we can address it in separate if we want to restore the feature right away.


kylebarron:
Thank you for this! I've got a busy day today and tomorrow and then am OOO Friday and Monday. I'll take a look at this next week!

kylebarron:
I added a separate `show_side_panel` argument to `Map`, so that users can separately decide whether to show the side panel or the tooltip.

Now with `show_tooltip=True, show_side_panel=False` we can get the old behavior back:

```py
m = Map(layer, _height=800, show_tooltip=True, show_side_panel=False)
with sidecar:
    display(m)
```

image

kylebarron (approved):
Thanks!

Issue #698, opened 2024-10-31
[BUG] `lonboard` conflicts with most sphinx themes


keewis:
## Context

In xarray-contrib/xdggs#80 I've tried to build `sphinx`-based documentation that includes notebooks using `lonboard` maps using `myst-nb` (but `nbsphinx` has the same issue).

This turns out to mess up the text styles for most of the `sphinx` themes (the ones I tested were `sphinx_book_theme`, `pydata_sphinx_theme` and `furo`).

See [the PR preview](https://xdggs--80.org.readthedocs.build/en/80/notebooks/tutorials/h3.html) for how exactly this breaks.

I'm not sure if this is an issue with `lonboard` / its dependencies, or with the `sphinx` themes, though.

## Environment

- OS: ubuntu 24.04
- Browser: firefox 131.0.3 and chromium 130.0.6723.69
- Lonboard Version: 0.10.3

## Steps to reproduce the bug

1. Setup environment and install the dependencies
2. Put `conf.py`, `index.md`, and the `air-traffic-control` notebook from the `lonboard` examples (might need to adjust the notebook kernel first, though) in a directory
3. run `python -m sphinx -M html -d _build/doctrees -W --keep-going -Ea -T -n . _build/html`
4. Compare `_build/html/air-traffic-control.html` with the one from the example in `lonboard`'s documentation.

<details><summary><tt>index.md</tt></summary>

````markdown
# Document Title

```{toctree}
air-traffic-control
```
````

</details>
<details><summary><tt>conf.py</tt></summary>

```python
extensions = ["myst_nb"]
root_doc = "index"
html_theme = "sphinx_book_theme"
```

</details>

<details><summary>Dependencies</summary>

```
python                    3.12.7          hc5c86c4_0_cpython    conda-forge
sphinx                    8.1.3              pyhd8ed1ab_0    conda-forge
myst-nb                   1.1.2              pyhd8ed1ab_0    conda-forge
myst-parser               4.0.0              pyhd8ed1ab_0    conda-forge
lonboard                  0.10.3             pyh5bfe37b_0    conda-forge
sphinx-book-theme         1.1.3              pyhd8ed1ab_0    conda-forge
pydata-sphinx-theme       0.15.4             pyhd8ed1ab_0    conda-forge
furo                      2024.8.6           pyhd8ed1ab_1    conda-forge
```

</details>

kylebarron:
[...]

Issue #696, opened 2024-10-29
[BUG] A short yet descriptive title for the bug you're reporting


deanm0000:
## Context

When using lonboard in VSC jupyter interactive in dark mode, once I display a Map, the output for everything carries the theme of the Map of white background. Before running a map it looks like: 
![image](https://github.com/user-attachments/assets/269cc5fb-2368-4eb0-95fa-b06f5a35fada)

## Resulting behaviour, error message or logs

After having run a map it looks like this:

![image](https://github.com/user-attachments/assets/775c43af-c9b4-4bca-b459-5b523f436cba)

## Environment

Platform:            Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python:              3.12.2 (main, Feb 25 2024, 16:35:05) [GCC 11.4.0]
- Lonboard Version: '0.10.3'
VSC:                   1.93.1
Chromium:        124.0.6367.243

## Steps to reproduce the bug

Using VSC with dark theme. Make a map then all cells get the Map's style, at least for background and font.

kylebarron:
I don't particularly understand CSS styling and what's happening here. But it's also quite difficult to provide CSS that works as expected in _every_ environment, when Jupyter/JupyterLab/Colab/vscode etc are all implemented in different ways. Happy to accept a PR or maybe @vgeorge has time at some point to look into this. (but lower priority than some bugs in the last release like https://github.com/developmentseed/lonboard/issues/675)

eKerney:
Hello @kylebarron and @vgeorge I'm experiencing a similar issue with VSC and am wondering if there is any room to make a contribution on this?  WSL2 ubuntu 22.04 - lonboard 0.10.3 - python 3.11.10

eKerney:
@kylebarron I haven't found a fix yet, though it is possible that this is an issue with VSCode and ipywidgets, and not necessarily lonboard - See this open ticket from the vscode-jupyter repo
#### [Add support for theming of IPyWidgets #7161](https://github.com/Microsoft/vscode-jupyter/issues/7161)

bjornasm:
This does not work with me, anyone else got it to work with lonboard? https://github.com/microsoft/vscode-jupyter/issues/7161#issuecomment-1616627670

kylebarron:
I _think_ this is closed via https://github.com/developmentseed/lonboard/pull/740

eKerney:
Great to see this fixed for VSCode!  
[...]

PR #740, opened 2025-01-29
Scope tailwind preflight styles

Bjoern-Rapp:
Anywidget loads CSS from the _css attribute to the global scope. A widget most take care not to conflict with any other CSS selectors. The stylesheet from the tailwind preflight, loaded with `@tailwind base`, will conflict with stylesheets from Sphinx themes (#698) or the VScode interactive display (#696).

To scope the preflight stylesheet to descendants of a wrapper should fix that.


kylebarron:
Oo interesting; so this would solve https://github.com/developmentseed/lonboard/issues/696?

@vgeorge any thoughts?

Bjoern-Rapp:
Yes, that will fix the problem with "leaking" CSS for both #696 and #698. 
#696 don't just affects  the widget, but the whole notebook/interactive window.

| Existing | PR | 
|----------|----|
| image |  image
 
It will not fix the unrelated problem/ feature request of adhering to dark or light themes inside the widget itself.

vgeorge:
@Bjoern-Rapp this is working for me. I tested in vscode and it doesn't seem to be leaking styles anymore. Would you be able to run `npm run prettier` on your branch and commit the results? This should fix the failing checks.

src/globals.css
```diff
@@ -1,3 +1,32 @@
...
+    Cantarell,
+    Noto Sans,
+    sans-serif,
+    "Apple Color Emoji",
+    "Segoe UI Emoji",
+    Segoe UI Symbol,
+    "Noto Color Emoji";
+  font-feature-settings: normal;
+  font-variation-settings: normal;
+  -webkit-tap-highlight-color: transparent;
```
vgeorge, line 26: @Bjoern-Rapp it didn't notice this when I reviewed yesterday. Could you explain why this was added?

Bjoern-Rapp: Some selectors in the preflight stylesheet targets html, body and :root elements. They become useless when scoped to the div.lonboard wrapper, since there are no html or body elements inside the wrapper. The only alternative then is to copy / duplicate those properties from the resulting stylesheet and apply them directly to the wrapper.

vgeorge:
@Bjoern-Rapp @kylebarron the background color change appears to be coming from NextUI, not Tailwind. Disabling [this line](https://github.com/developmentseed/lonboard/blob/1474a9d9f67a93fbbbd188487debb3870d0909e4/tailwind.config.js#L13), at least in the VSCode interactive code display, keeps the existing color. Adding overrides to the global styles might work for VSCode, but I would be in favor of a more permanent solution that avoids maintaining the NextUI-related overrides. Currently, we are using only two components from NextUI, Button and Tooltip. I believe we should consider replacing them with custom components and dropping the NextUI dependency as it might conflict with other environments. If you agree we can track this in a separated ticket.

Bjoern-Rapp:
The change in background colour change comes from NextUI, but Tailwind still comes with other changes to the global style that may be unwanted in other environments than VScode, such as Sphinx, if the "base" stylesheet is not scoped in some way. The stylesheet still contains changes to the font family, even when NextUI is removed.
Removing NextUI will reduce the size of the final stylesheet, but isn't really solving the issue.

This solutions should work in all environments, not just VSCode, but there is probably better ways to scope the styles,
or perhaps a way to automatically re-target selectors that targets html, body and :root when scoped?

kylebarron:
Thank you!

PR #745, opened 2025-02-05
Support Py3.12 in tests by constraining pyproj in lockfile to versions with wheels

ceholden:
This PR adds environment markers to the `uv` tooling configuration [constraint-dependencies](https://docs.astral.sh/uv/reference/settings/#constraint-dependencies) for `pyproj`, allowing us to have a lockfile with `pyproj` wheels for Python versions we support without restricting versions for downstream consumers.

I believe the path to supporting 3.13 in tests will require updating `fiona<1.10` to something like `fiona>1,<=1.10`. It also looks safe to bump DuckDB in our lockfile but we will have to invalidate the CI cache so it doesn't point to extensions compiled for `1.1.3`


kylebarron:
Thank you! I think it's fine to add only 3.12 to CI; 3.13 might be more trouble than it's worth

ceholden:
> Thank you! I think it's fine to add only 3.12 to CI; 3.13 might be more trouble than it's worth

+1 I'll remove that for now. It looks like the install is choking on not having a wheel for `fiona<1.10`, but there are wheels available if we can relax that to support `1.10.x`

ceholden:
Relocking bumped DuckDB to 1.2 and I think there's some caching of DuckDB extensions going on that caused the last CI failures. I undid the bump in `uv.lock` for DuckDB for now to check if that's true

kylebarron:
Thank you!

PR #753, opened 2025-02-11
Enable more ruff linters

kylebarron:
There are still some failing lints. Tomorrow we can fix a few more but mostly relax lints to be able to merge.


kylebarron:
Using the ruff action gets annotations on the PR 
image

PR #769, opened 2025-03-14
Sanitize geomcolname

### LINKED ISSUE #768: [BUG] duckdb `geom_col_name` not actually sanity checked
## Context

I am calling `viz` on a duckdb query where I did not explicitly add a column name, so the implicit column name contains unexpected characters such as "(" (which are actually valid in a duckdb table column name, but not desired in the context of lonboard given the intent of [this line](https://github.com/developmentseed/lonboard/blob/98d68f750ece7d09e209edba354e124ceb3416d5/lonboard/_geoarrow/_duckdb.py#L117))

```python
viz(duckdb.sql("""select ST_GeomFromText('POINT (1 2)')"""))
```

Now tihs _will_ fail, but we should get a meaningful error (or visualize it correctly anyway -- probably less good of an idea).

**Describe what happened:**
```python
!pip install lonboard --quiet

import duckdb

from lonboard import viz

duckdb.sql("""INSTALL spatial; LOAD spatial;""")

print(duckdb.sql("""select ST_GeomFromText('POINT (1 2)')"""))
# Returns:
# ┌────────────────────────────────┐
# │ st_geomfromtext('POINT (1 2)') │
# │            geometry            │
# ├────────────────────────────────┤
# │ POINT (1 2)                    │
# └────────────────────────────────┘



viz(duckdb.sql("""select ST_GeomFromText('POINT (1 2)')"""))
# throws duckdb.ParserException: Parser Error: syntax error at or near "("

viz(duckdb.sql("""select ST_GeomFromText('POINT (1 2)') as geom"""))
# opens widget
```

## Environment

- OS/Browser: Google Colab
- Lonboard Version: 0.10.4

## Steps to reproduce the bug

see above

danielsparing:
## What I am changing

Fix sanity check of `geom_col_name` to actually check the whole string.

## How I did it

`re.match(re_match)` was matching any string containing _any_ `[a-zA-Z]` character, regardless of the rest.
The solution is `re.fullmatch()`. Another way would have been to append `r"^...$"` to the beginning and end of `re_match`.

## How you can test it
Wrote test `test_sanitize_column_name()`.

## Related Issues
Closes #768

In general, the issue of [hardcoded-sql-expression](https://github.com/developmentseed/lonboard/blob/98d68f750ece7d09e209edba354e124ceb3416d5/lonboard/_geoarrow/_duckdb.py#L1) is still out there.

### DIFF
--- a/lonboard/_geoarrow/_duckdb.py
+++ b/lonboard/_geoarrow/_duckdb.py
@@ -115,7 +115,7 @@ def _from_geometry(
     # A poor-man's string interpolation check
     # We can't pass in SQL-templated strings for the column name
     re_match = r"[a-zA-Z][a-zA-Z0-9_]*"
-    assert re.match(
+    assert re.fullmatch(
         re_match,
         geom_col_name,
     ), f"Expected geometry column name to match regex: {re_match}"
--- a/tests/test_duckdb.py
+++ b/tests/test_duckdb.py
@@ -240,3 +240,25 @@ def test_geometry_only_column_type_geometry():
 
     # Should create layer without erroring
     _layer = ScatterplotLayer.from_duckdb(query, con)
+
+
+def test_sanitize_column_name():
+    # For WKB parsing
+    pytest.importorskip("shapely")
+
+    con = duckdb.connect()
+    sql = f"""
+        INSTALL spatial;
+        LOAD spatial;
+        SELECT
+          * EXCLUDE geom,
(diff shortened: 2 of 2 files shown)


kylebarron:
Oh those CI failures are unrelated because we don't pin a single version of ruff. I'll bump the ruff version in a separate PR

kylebarron:
Thanks!

PR #778, opened 2025-03-28
Reduce ruff warnings

danielsparing:
nit: minor changes to clean up ruff / pre-commit output


docs/CHANGELOG.md
```diff
@@ -1 +1 @@
-../CHANGELOG.md
\ No newline at end of file
+../CHANGELOG.md
```
kylebarron, line 1: What changed here?

pyproject.toml
```diff
@@ -119,6 +119,8 @@ ignore = [
     "B028",    # No explicit `stacklevel` keyword argument found
     "B011",    # Do not `assert False` (`python -O` removes these calls), raise `AssertionError()`
     "D100",    # Missing docstring in public module
+    "D203",    # incorrect-blank-line-before-class
+    "D213",    # multi-line-summary-second-line
```
kylebarron, line 123: Were there previously errors that these were catching? Main is clean in terms of lints?

danielsparing: So these two were already ignored, because they are conficting two other rules (and hence _would_ catch "errors" if they were active), as the below warning within the CI runs states (e.g. [here](https://github.com/developmentseed/lonboard/actions/runs/14136529586/job/39609473134#step:13:24)):

```
  - 'extend-per-file-ignores' -> 'lint.extend-per-file-ignores'
warning: `incorrect-blank-line-before-class` (D203) and `no-blank-line-before-class` (D211) are incompatible. Ignoring `incorrect-blank-line-before-class`.
warning: `multi-line-summary-first-line` (D212) and `multi-line-summary-second-line` (D213) are incompatible. Ignoring `multi-line-summary-second-line`.
```

Therefore it comes down to your preference:
- Do you want an extra newline between class definition and docstring? ([D203](https://docs.astral.sh/ruff/rules/incorrect-blank-line-before-class/) vs [D211](https://docs.astral.sh/ruff/rules/blank-line-before-class/))
- Within the docstring, do you want the summary line to start on the first or second line? ([D212](https://docs.astral.sh/ruff/rules/multi-line-summary-first-line/) vs [D213](https://docs.astral.sh/ruff/rules/multi-line-summary-second-line/))

In this commit I made the choices so that the existing codebase still passes. If you were to invert either or both of them, then that would require an (automatic) edit all the docstrings elsewhere too.

kylebarron: Oh... I see.

- Prefer no extra newline between class def and docstring
- Prefer summary line on the first line (d212)

danielsparing: Then I think it's all OK:
- the repo is already following these preferences
- D203 and D213 _would have_ gone against these preferences -- but they were already ignored as they were conflicting with their opposite rules -- but now with this PR, there won't even be a warning emitted about this clash.

kylebarron:
Thank you!

PR #777, opened 2025-03-28
Switch to the DuckDB Expression API to mitigate sql injection risk

danielsparing:
## Available PR templates

- [Default](?expand=1&template=default.md)
- [Version Release](?expand=1&template=version_release.md)
- _Alternatively delete and start empty_

PR #784, opened 2025-04-19
Remove references to custom connection parameter

### LINKED ISSUE #1: Separate into multiple widgets/layers?

danielsparing:
## What I am changing

- Remove doc references to the (now unused) `con` parameter in `from_duckdb` and `viz`.
- Added new section on `str` input to `from_duckdb`, as now _that_ is a main case for `con`.

## How I did it

- n/a

## How you can test it

- n/a

## Related Issues

- Following https://github.com/developmentseed/lonboard/pull/777 which earlier removed this parameter


danielsparing:
@kylebarron the last `Implementation` section is admittedly now a bit open-ended, as the argumentation doesn't lead to the conclusion to use the `con`. I'll leave it to you to cut it entirely, or maybe leave it in some form still, as we're still calling `st_aswkb`.

docs/ecosystem/duckdb.md
```diff
@@ -54,50 +54,23 @@ m
...
 
 ```py
 import duckdb
 from lonboard import viz
 
-con = duckdb.connect()
-con.execute("CREATE TABLE spatial_table AS ...;")
-viz(con.table(), con=con)
+duckdb.sql("CREATE TABLE spatial_table AS ...;")
+viz(duckdb.table("spatial_table"), con=con)
```
kylebarron, line 64: There's still a `con=con` here. Is that supposed to be taken out?

docs/ecosystem/duckdb.md
```diff
@@ -113,32 +86,10 @@ m = Map(layer)
...
 
 We could require that users always call `ST_AsWKB` on the geometry column on every query, but this is unwieldy and easy to forget. Instead, by requiring that the user pass in a `DuckDBPyRelation` object instead of a `DuckDBPyConnection`, we're able to use the [replacement scan feature](https://duckdb.org/docs/api/python/relational_api#sql-queries) of `DuckDBPyRelation` to call _back_ into DuckDB and perform our _own_ `ST_AsWKB` call on the data, before it's left DuckDB memory, and then export the WKB geometries out to Arrow.
-
-This is why we sometimes need access to the `DuckDBPyConnection` (the `con` variable) that created the `DuckDBPyRelation`, because we need to perform a new query on the same connection to convert the geometry column into WKB.
-
-Note that geometry types other than `GEOMETRY`, namely `WKB_BLOB`, `POINT_2D`, `LINESTRING_2D`, `POLYGON_2D`, and `BOX_2D`, do not suffer from this limitation. We can directly interpret those types in Python and do not need to call back into DuckDB.
```
kylebarron, line 144: I think this last sentence can stay in.

docs/ecosystem/duckdb.md
danielsparing: Ah oops yes!

docs/ecosystem/duckdb.md
danielsparing: Yess true

kylebarron:
Thanks!

PR #789, opened 2025-05-04
updates for polygon layer documentation

### LINKED ISSUE #788: PolygonLayer get_line_color docstring has incorrect information
The docstring for PolygonLayer get_line_color indicates that it [only applies if extruded is True](https://github.com/developmentseed/lonboard/blob/229e995d2597b13035692381beedd094ef947f70/lonboard/_layer.py#L1213C1-L1214C1), but it should be if `stroked` is True.  I'll put in a PR to fix it!

ATL2001:
fixes #788 and changes 'line' to 'outline' to make what line we're talking about more clear/consistent with other use of 'outline'


kylebarron (approved):
Thanks!

PR #811, opened 2025-06-18
change duckdb import to lazy

### LINKED ISSUE #810: [BUG] duckdb now became a hard dependency
## Context

Context is installing lonboard in an environment where duckdb is not yet installed, for example [Databricks serverless environment version 2](https://docs.databricks.com/aws/en/release-notes/serverless/environment-version/two) (Google Colab has duckdb installed).

This is relevant in case the user would not want to use `duckdb`.

**Describe what happened:**
```
%pip install lonboard
```
```
import lonboard
# Outputs:
# ModuleNotFoundError: No module named 'duckdb'
```

## Environment

- OS: Ubuntu 22.04.5 LTS
- Browser: Chrome
- Lonboard Version: 0.11.0

## Steps to reproduce the bug

Describe the actions that led you to encounter the bug. Example:

1. make sure duckdb is NOT installed, so `import duckdb` fails
2. try and import lonboard

Most probably an unintended consequence of https://github.com/developmentseed/lonboard/pull/777 (sorry! :) )

danielsparing:
## What I am changing
Remove duckdb imports outside TYPE_CHECKING and move into the relevant function as lazy import

## How I did it
I realize now that the imports outside TYPE_CHECKING lead to an error at `import lonboard`-time if the should-be-optional prerequisite (duckdb) is not installed

## How you can test it
Not sure, as the tests did not catch the earlier issue

## Related Issues
Closes #810 

(or alternatively, duckdb could become an actual dependency that's always installed with?)


danielsparing:
sorry, closed in favor of a cleaner #813

PR #813, opened 2025-06-18
fix: move duckdb imports to inside function

### LINKED ISSUE #810: [BUG] duckdb now became a hard dependency

danielsparing:
to avoid duckdb becoming a hard dependency

## What I am changing / How I did it
move import to inside function

## How you can test it
might trigger a ruff rule around imports not at top of file

## Related Issues
closes #810 

(or alternatively, duckdb should explicitly become a hard dependency)

### DIFF
--- a/lonboard/_geoarrow/_duckdb.py
+++ b/lonboard/_geoarrow/_duckdb.py
@@ -14,7 +14,6 @@ from arro3.core import (
     list_array,
     struct_field,
 )
-from duckdb import ColumnExpression, FunctionExpression
 
 from lonboard._constants import EXTENSION_NAME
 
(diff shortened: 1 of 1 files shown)


kylebarron:
Thanks; looks like that was an unintended consequence from https://github.com/developmentseed/lonboard/pull/777

Issue #781, opened 2025-03-31
Include _height in API docs


kylebarron:
Looks like it's not included

ATL2001:
do you know why it's `_height` and not just `height`? it looks like there is a docstring in there  ["""Height of the map in pixels.](https://github.com/developmentseed/lonboard/blob/0150ae8d2f76e899763844dad2b90e4688001994/lonboard/_map.py#L181)

I don't have a lot of experience with mkdocs, but in sphinx docs the autodocs it creates will skip anything with the leading underscore(although I think the skipping can be overridden), and I'd wager a guess its the same here.

kylebarron:
It's `_height` because this parameter is a total hack. Ideally we shouldn't need to force a specific height, and instead should let the widget take up the entire space of its container, but I don't know how to do that.

ATL2001:
I wasn't exactly looking for a solution to this, but I just stumbled on the documentation of mkdocstrings-python that does confirm that it is being filtered out by the leading underscore: https://mkdocstrings.github.io/python/usage/configuration/members/#filters

> The default value (["!^_[^_]"]) means: render every object, except those starting with one underscore, unless they start with two underscores. It means that an object whose name is hello, __hello, or __hello__ will be rendered, but not one whose name is _hello.

unfortunately it looks like changing that would require a mkdocstrings sponsorship, so unless devseed is a sponsor, it appears `_height` would need to be renamed to `height` for it to show up in the docs.

kylebarron:
No, you can change that config value without needing the members-only build.

kylebarron:
I spent a little time on this and can't get `_height` to display. I have no idea what's going wrong.

ATL2001:
I can't promise anything right now, but would you be open to renaming it to `height` if I can get the map to fill the whole container?  (seeing as that's the desired behavior anyway, and I'm fairly confident if it didn't have that leading underscore it would show up) 

full disclaimer, I'm not positive I can fix up the JS/HTML end of things or that I'll have time to work on it in the next few days, but I'll at least try to give it a shot at some point if you think it's worthwhile

kylebarron:
I would love for the height to automatically fill the entire container. That's essentially why `_height` is private, because it was always the goal to be unnecessary. I don't have the time or the JS knowledge to fix it, especially because different environments tend to work differently.

ATL2001:
[...]

PR #816, opened 2025-06-22
feat: replace map `_height` with `height`

ATL2001:
per what we discussed in #781 I believe I've come up with a good solution here.  I had to add some CSS to the data-overlay-container which appears to be generated by the typescript which the lonboard div is housed in. and then I needed to make some tweaks to the map div which lives inside the lonboard one as well.  then I changed `_height` to `height` and tweaked the way the default was setup so as to not make the default functionality of the creation of the map change.

but now if  the user knows they want to put the map into an IFrame or some other container that has a fixed height, they can specify `height="100%"` and the map will fill the container's height.

I also checked in an example notebook so you can check the behavior yourself.  I do not intend to keep that in there (unless you think it would be a benefit for others, if so I should probably retool it to make it a little nicer)

If this looks good let me know and I can find some time to update the docs so they match the behavior 👍


lonboard/_map.py
```diff
@@ -177,7 +179,7 @@ def on_click(self, callback: Callable, *, remove: bool = False) -> None:
     Indicates if a click handler has been registered.
     """
 
-    _height = t.Int(default_value=DEFAULT_HEIGHT, allow_none=True).tag(sync=True)
+    height = t.Unicode(default_value=DEFAULT_HEIGHT).tag(sync=True)
```
kylebarron: I don't really want to have a breaking change here, so if we allow strings here, it should be a `t.Union` of strings and integers (or at least we need to validate that we can pass in an integer and it'll get converted to a string, but we probably want to add a `px` suffix?)

kylebarron:
It would be really nice if the HTML export automatically filled the entire window: 

```py
Map([]).to_html("test.html")
```
gives:
image

ATL2001:
oh yeah, this exporting as height 100% looks really sharp! you want a map, you get a map! 💥

lonboard/_map.py
```diff
@@ -567,16 +567,22 @@ def to_html(
...
-                title=title or "Lonboard export",
-                template=_HTML_TEMPLATE,
-                drop_defaults=False,
-                # Necessary to pass the state of _this_ specific map. Otherwise, the
-                # state of all known widgets will be included, ballooning the file size.
-                state=dependency_state((self), drop_defaults=False),
-            )
+            original_height = self.height
+            try:
+                self.height = "100%"
```
kylebarron: Will this cause the JS side to be janky? Is there a way to set this without causing a JS side re-render? 

Like maybe use `hold_trait_notifications` to avoid sending any new events while temporarily setting `.height`? https://traitlets.readthedocs.io/en/stable/using_traitlets.html#holding-trait-cross-validation-and-notifications

src/index.tsx
```diff
@@ -214,8 +214,14 @@ function App() {
     [isOnMapHoverEventEnabled, justClicked],
   );
 
+  let height = String(mapHeight);
+  if (typeof mapHeight === "number") {
+    // if a number is used for height, assume pixels
+    height = height + "px";
+  }
```
kylebarron: I was thinking of suggesting this validation be done on the Python side, so the Python side is always sending a string to the JS side. There are several places that we already use custom validators

lonboard/_map.py
ATL2001: good call, I didn't notice that re-rendering happening

kylebarron:
To fix CI you can run `uv run --group dev pre-commit run --all-files` and then commit the output. The formatter is failing on the notebook

ATL2001:
I was planning on deleting that notebook eventually, but i went ahead and fixed it up, but I'm not sure what this new error means :(

kylebarron:
If you look at the git diff, you did something to change the symlinks. And so the docs website build isn't able to find the `index.md` (pointing to the top-level README)

ATL2001:
weird, that `uv run --group dev pre-commit run --all-files` added blank newlines at the end of the .md files which seems to have broken the docs somehow (I wonder if it's somehow related to me working on this from my windows machine?) I deleted them and all the checks passed now though! 🎉 

when I run that pre-commit this is what it tells me:
{7C2159E8-1980-4465-B6F3-F269188C1BAF}

kylebarron:
Because of #820, I had to rewrite history to remove 140MB of accidentally-committed files at the start of the project. So this PR is now all messed up; apologies for that. Maybe it would be best to start a new PR from scratch 🫣 🙈

ATL2001:
I don’t have my computer this week but should have some time in the evenings next week to get to this

kylebarron:
No worry, I should be able to recreate the PR too. I should've realized the git history rewrite  would mess up WIP PRs. 

I don't think I checked the last commit on this branch to see if it seemed functional with all our notebook targets. It sadly requires manual testing to verify if it works with all backends

kylebarron:
I took a quick look but I don't think my git skills are good enough to figure out how to rebase this PR onto the current main. Somehow we only want to keep the last 10 commits or so

kylebarron:
Closing in favor of #835

Issue #706, opened 2024-11-13
Use `juv` for improved reproducibility of examples


kylebarron:
https://github.com/manzt/juv

PR #821, opened 2025-07-10
feat: Migrate notebook examples to dependency management using `juv`

kylebarron:
In-progress conversion of notebooks to switch to using `juv` instead of 

Should probably suggest using `uvx juv run notebook` so that users don't have to install `juv`, they only have to install `uv`

Closes https://github.com/developmentseed/lonboard/issues/706

juv is so nice! Thanks @manzt!

### DIFF
--- a/examples/migration.ipynb
+++ b/examples/migration.ipynb
@@ -1,6 +1,32 @@
 {
  "cells": [
   {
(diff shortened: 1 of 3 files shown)


manzt:
uvx juv all the way!

Issue #729, opened 2025-01-08
[BUG] Map height fixed at 24rem


wx4stg:
When embedding lonboard in panel (and probably other similar toolkits), the map height is fixed to 24rem, regardless of the parent container's height.

This issue is new since 0.9.3, I believe this issue was first introduced in 44a45f8. Continuing discussion from #264 

## Environment

- OS: Fedora 41
- Browser: Chrome/Brave 
- Lonboard Version: 0.10.3

## Steps to reproduce the bug

```python
import panel as pn
from lonboard import Map
from lonboard.basemap import CartoBasemap

pn.extension()

title = pn.pane.Markdown('# THIS IS A TITLE')
lonb_map = Map([], basemap_style=CartoBasemap.Positron)
lonb_map_panel_container = pn.pane.IPyWidget(lonb_map, sizing_mode='stretch_both', align='center')

pn.Column(title, lonb_map_panel_container).servable()
```

![Image](https://github.com/user-attachments/assets/1a6ce29c-0acf-4b5d-a129-8661d2d91067)

wx4stg:
I would be happy to *try* to troubleshoot this myself, to do this I could just (1) clone the repo (2) make my modifications (3) follow the [python/uv instructions](https://github.com/developmentseed/lonboard/blob/main/DEVELOP.md#python) for compiling/running? ...or do I also need to do the [javascript/node](https://github.com/developmentseed/lonboard/blob/main/DEVELOP.md#javascript) steps?

*Just to be completely clear, I'm finishing up my MS right now and will likely not be able to start on this until later this summer... This came up in a discussion I had this week so I'm commenting here so I don't forget...*

kylebarron:
Yes the [develop docs](https://github.com/developmentseed/lonboard/blob/main/DEVELOP.md) should have everything you need.

You'll need to follow the JS steps as well if you want to test locally, since there's a JS build step needed for the Python package.

danielsparing:
@wx4stg I think you can pass `_height` to `Map()` such as `Map( ... _height=1000)`.

But I am running into the same myself in two other, related cases:
- `viz()` doesn't accept a height parameter
- when exporting a HTML (or inserting into another container, as @wx4stg 's example), you might want to use "100%", which is not possible here, and was once already implemented in https://github.com/developmentseed/lonboard/pull/377
- the `_height` parameter carries the note "This API is not yet stabilized and may change in the future." so I guess we shouldn't rely on it too much :)

[...]

PR #835, opened 2025-07-20
feat: Use `100%` as default CSS height

### LINKED ISSUE #781: Include _height in API docs

ATL2001:
This PR will replace the existing private `_height` property of a Lonboard map with a new public `height` property.  The new property will accept an integer (assumed to be the number of pixels) for the map, or a string that contains a CSS height (for example, '100%' or '24rem') 

When lonboard maps are exported to an HTML file using the `.to_html()` method, the map in the resulting HTML file will always be set to 100% height so the map will fill the entire browser window. 

I believe this will resolve #781 since the new property doesn't have the preceding underscore. as well as #729

### DIFF
--- a/lonboard/types/map.py
+++ b/lonboard/types/map.py
@@ -15,7 +15,7 @@ if TYPE_CHECKING:
 class MapKwargs(TypedDict, total=False):
     """Kwargs to pass into map constructor."""
 
-    _height: int
+    height: int | str
     basemap_style: str | CartoBasemap
     parameters: dict[str, Any]
     picking_radius: int


kylebarron:
Do you know if there's a way to default to 100%?

In JupyterLab in the sidecar widget it doesn't expand all the way:
image

ATL2001:
hmmm, I'll check into it tonight to see if i can make that work

ATL2001:
Yes! I looked at sidecar's docs, and in them they are showing a ipyleaflet map, so I jumped over there and found that they're [setting the min-height on their maps](https://github.com/jupyter-widgets/ipyleaflet/blob/290076e0b8b7d2a4056f9472592be5a8f66ed949/python/jupyter_leaflet/css/jupyter-leaflet.css#L16C1-L19C2) so I tried that and I think we've got a winner! 

We can still set the height to a number (for pixels) or to a string of a valid CSS height and then the map will show as that height in the normal jupyterlab output 

image

image

kylebarron:
This is amazing! Thank you! This looks great on JupyterLab!
image


However it looks like it's not rendering at all in vscode:
image

It's tricky, but I think we ideally want to verify in:

- [x] JupyterLab
- [x] JupyterLab sidecar
- [x] Jupyter Notebook
- [x] Colab
- [x] Vscode Jupyter Console
- [x] Vscode Jupyter Notebook
- [x] HTML output

src/index.tsx
```diff
@@ -215,11 +215,11 @@ function App() {
   );
 
   return (
-    <div className="lonboard">
+    <div className="lonboard" style={{ minHeight: mapHeight, height: "100%" }}>
       <div
         id={`map-${mapId}`}
         className="flex"
-        style={{ height: mapHeight ? `${mapHeight}px` : "24rem" }}
+        style={{ width: "100%", height: "100%" }}
```
vgeorge: @ATL2001 @kylebarron This approach will not work in every environment because `<Map>` requires the enclosing `<div>` to have a fixed height, not 100%, which is why we are setting the `mapHeight` here.

ATL2001: @vgeorge what are your thoughts if we switch the minHeight to be '100%' and the height to come from mapHeight?  that makes it show up in jupyterlab, jupyterlab sidecar and in an .ipynb in VScode 

(I'm honestly not sure what jupyter console in Vscode is, I can't seem to get lonboard to show a map in an oldschool jupyter notebook, and I've never used colab 😞 , so I didn't check any of those) 

image

kylebarron: > jupyter console in Vscode

That's this: https://code.visualstudio.com/docs/python/jupyter-support-py

image

(I do 99% of my Python development in it)

kylebarron:
I verified that this works in Colab.

(You can follow the release CI:
https://github.com/developmentseed/lonboard/blob/a002d7fe0d3e12513ad5b3f57900846a945f5eb0/.github/workflows/release.yml#L33-L41

```
npm run build
python -m build
```

Then upload the generated `.whl` into Colab and `%pip install` from the local `.whl` file.

image

kylebarron:
Verified that this works in Vscode Jupyter Console:

image

kylebarron:
Verified that this works in Vscode Jupyter notebooks:

image

kylebarron:
Validated in JupyterLab: 
image

kylebarron:
Validated in JupyterLab sidecar:
image

kylebarron:
Validated in HTML output: 
image

kylebarron:
Validated in jupyter notebook: 
image

kylebarron (approved):

kylebarron:
Since I was able to validate output in every target environment, I'm 👍 on merging this. @vgeorge any hesitations?

vgeorge:
@kylebarron @ATL2001 this is great, really glad you found a working solution across all environments. 🎉

Issue #494, opened 2024-05-01
Allow dict input for Map Style


kylebarron:
The dict should be a valid mapbox style spec object.

We might also want to have helpers for a single-layer raster-tile basemap. E.g. `lonboard.basemap.RasterBasemap` which takes in an xyz url. We should have an example with standard OSM tiles. 

There are a bunch of default public raster tile basemaps [here](https://leaflet-extras.github.io/leaflet-providers/preview/) (referenced from [here](https://python-visualization.github.io/folium/latest/getting_started.html#Choosing-a-tileset)). Also see https://github.com/geopandas/xyzservices, which may be a better integration point. We should be able to integrate directly with the [`TileProvider` class](https://xyzservices.readthedocs.io/en/stable/api.html#xyzservices.TileProvider).

kylebarron:
It may also be useful to refer to these basemaps as `MaplibreBasemap` objects, where we could also have `DeckglBasemap` objects in the future, for applications that don't want to render Maplibre beneath deck.gl

PR #860, opened 2025-08-25
feat: geopandas extension; explore method

knaaptime:
supercedes #728


lonboard/geopandas.py
```diff
@@ -0,0 +1,310 @@
...
+    cat_codes = cat_codes.cat.rename_categories({-1: len(cat_codes.unique()) - 1})
+    unique_cats = categories.dropna().unique()
+    n_cats = len(unique_cats)
+    colors = colormaps[cmap].resampled(n_cats)(list(range(n_cats)), alpha, bytes=True)
+    colors = np.vstack([colors, nan_color])
+    temp_cmap = dict(zip(range(n_cats + 1), colors, strict=True))
+    return apply_categorical_cmap(cat_codes, temp_cmap)
+
+
+def _query_name(name: str) -> CartoBasemap:
```
kylebarron: I'd prefer just depending on `xyzservices` for the geopandas integration:

https://github.com/geopandas/geopandas/blob/4f9f361a2ba94db4c02e1fb5b6b2b09abf4278e6/geopandas/explore.py#L373-L386

I don't think we should implement our own mapping. After all, the cartodb basemap styles are defined in xyzservices too, right?

lonboard/geopandas.py
```diff
@@ -0,0 +1,310 @@
...
+    "CartoDB Darkmatter": CartoBasemap.DarkMatter,
+    "CartoDB Darkmatter No Label": CartoBasemap.DarkMatterNoLabels,
+    "CartoDB Voyager": CartoBasemap.Voyager,
+    "CartoDB Voyager No Label": CartoBasemap.VoyagerNoLabels,
+}
+# Convert keys to lower case without spaces
+_BASEMAP_PROVIDERS = {
+    k.translate(_QUERY_NAME_TRANSLATION).lower(): v
+    for k, v in _basemap_providers.items()
+}
```
kylebarron, line 40: So by depending on xyzservices here we should be able to remove all of this.

lonboard/geopandas.py
kylebarron: To be clear, I mean depending on it conditionally from this module, not as a required dependency of the package.

Also, soon, we'll support passing in an xyzservices object directly. See #494

lonboard/geopandas.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def __init__(self, pandas_obj: gpd.GeoDataFrame) -> None:
+        """Initialize geopandas extension."""
+        self._validate(pandas_obj)
+        self._obj = pandas_obj
+
+    @staticmethod
+    def _validate(obj: gpd.GeoDataFrame) -> None:
+        if not isinstance(obj, gpd.GeoDataFrame):
+            raise TypeError("Input Must be a geodataframe")
```
kylebarron, line 55: We can just put this in `__init__` instead of making a separate static function that's only called once.

lonboard/geopandas.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def explore(  # noqa: C901, PLR0912, PLR0913, PLR0915
+        self,
+        *,
```
kylebarron: I know this is the opposite of what I [said before](https://github.com/developmentseed/lonboard/pull/728#discussion_r2153144389), but I think we should remove `*` so that the signature is exactly the same as the upstream `GeoDataFrame.explore` (even though the thought of someone passing in all these parameters positionally is 😬 )

lonboard/geopandas.py
```diff
@@ -0,0 +1,310 @@
...
+        map_kwargs: MapKwargs | None = None,
+        classification_kwds: dict[str, str | IntFloat | ArrayLike | bool] | None = None,
+        nan_color: list[int] | NDArray[np.uint8] | None = None,
+        color: str | None = None,
+        vmin: float | None = None,
+        vmax: float | None = None,
+        wireframe: bool = False,
+        tiles: str | None = None,
+        highlight: bool = False,
+        m: Map | None = None,
```
kylebarron: I think it would be good to make this signature exactly the same as the upstream `geopandas.explore` signature. There are a bunch of differences:

- `k` exists in both but has a different default
- `color` exists in `geopandas.explore` but not here
- `elevation`, `elevation_scale`, etc. exist here but not in `geopandas.explore`.
- `scheme`, `k`, `highlight`, etc exist in both but have a different ordering

---

In particular, this gets to the _point_ of this integration and of this PR, which has been kinda unclear to me thus far. 

This API is essentially "a basket of arguments to do everything and anything", which in my opinion is awful API design. It's too easy to have complex interactions between arguments. It might take a few lines instead of one to make a map with Lonboard's standard APIs, but each layer and map object is very predictable with how they interact.

We already have a function in lonboard to take generic input (including GeoDataFrames) and plot them. That's `viz`. It's simple, straightforward, and extensible. I don't want to have another function like `viz` (at least, officially part of Lonboard) 

---

So that said, I think it's important that, if we want to merge this functionality _into Lonboard_, it should have **exactly the same API** as `GeoDataFrame.explore`, so that a user can change `GeoDataFrame.explore` to `GeoDataFrame.lb.explore` and it'll _always_ work out of the box.

Another possibility is talking to `geopandas` about natively implementing this there.

And besides that, it's also a fine option to have this as a standalone third-party Python package, `lonboard-geopandas`!

knaaptime: i can stick it somewhere else if you prefer. Martin said he'd prefer it here not in geopandas

>So that said, I think it's important that, if we want to merge this functionality into Lonboard, it should have exactly the same API as GeoDataFrame.explore, so that a user can change GeoDataFrame.explore to GeoDataFrame.lb.explore and it'll always work out of the box.

this was always my intent, except that the lonboard version would have some additional args to support lonboard's additional features (e.g. elevation, wireframe and scale). Given those additional features i'd expect people could *mostly* swap from the vanilla `explore` to this version, but not always.  I think the "ability to do anything and everything" really comes from the desire to match exactly the geopandas explore function. What lonboard *doesnt* have is the ability to use native classification schemes, which is a pain.
 
`color` exists in both (its line 76 here). Different defaults and ordering are simple fixes if you do want this here. Personally i think this would be much easier to have in lonboard itself for discoverability and less need for any additional package, especially since its only a little bit of code, but again, if you find it too burdensome i can find somewhere else

kylebarron: I think as a first implementation, I'd leave off any additional Lonboard-specific features and keep it exactly the same as the upstream GeoPandas signature, with the same ordering.

We should allow all the same parameters, and if one doesn't make sense in Lonboard, we should print a warning, but otherwise make sure something still renders.

lonboard/geopandas.py
knaaptime: ok, this was just following the instructions from the pandas [extension guidelines](https://pandas.pydata.org/docs/development/extending.html)

lonboard/geopandas.py
knaaptime: agree that would be ideal and ive never loved this mapping but there was no roadmap to supporting xyzservices when i started this a year ago

lonboard/geopandas.py
kylebarron: I'm still tepid about maintenance of this.

One idea is seeing about refactoring some of the upstream GeoPandas code, so that it isn't one big monolithic function. It would be nice if we could call `geopandas.explore._explore_prep` to manage a lot of the data prep that isn't related to the actual plotting in folium. Then we wouldn't have to duplicate all of the internals.

knaaptime: i get it; i'm fine with whatever you choose. I can all but guarantee geopandas has no bandwidth to look at their internals, but on the slight off-chance, i'll let @martinfleis weigh in. I can get rid of the lonboard-specific arguments if you prefer, though i do think having them available is pretty handy. One argument in favor of keeping them around is the vanilla `explore` already accepts catchall keywords, so it *would* always work to switch between functions, even with the extra params

Screenshot 2025-10-15 at 3 41 35 PM

knaaptime: another thing that's tough about trying to do the perfect 1:1 mapping with the existing explore function is that folium has a lot of stuff that works differently than lonboard. While a lot of the (admittedly ugly) code in this PR tries to handle those differences appropriately and stick stuff 'where it belongs', there would still be a lot of folium-specific things that don't make sense here, e.g. highlight, popup, attr, marker_type, marker_kwds, legend_kwds, highlight_kwds, style_kwds.

That is, lonboard doesn't have legends or attribution; 'marker_kwds' and 'style_kwds' are instead passed to their layer-specific analogs (like PolygonLayerKwargs); lonboard has side_panel versus folium's popup; sometimes the nomenclature is different (tooltip vs show_tooltip). In some cases, those are fairly easy to shoehorn, but in others (like side_panel vs popup) shooting for perfect parity probably is not an ideal goal. 

A small handful of related thoughts:

As both a user and an instructor, I'd find a lonboard-based `explore`-like function (that *mostly* works the same as the existing one--even if only for the classification schemes, which are indispensable for anything human geography-related) to be incredibly valuable, especially if it existed somewhere central like geopandas or lonboard. Indeed, I'd probably never bother with the vanilla explore anymore, given all the benefits of lonboard over folium. I'm confident that view would be shared widely. But if neither geopandas nor lonboard want that functionality, i'll find somewhere that's conveninent for me--but then only my students and a handful of colleagues will know where to find it and that feels like a net loss to the ecosystem. I'd be more than happy to contribute to ongoing maintenance, but the infrastructure overhead of putting together yet another package, just for this, probably isn't tenable. I doubt that changes anyone else's calculus, but just figured i'd lay it out.

lonboard/geopandas.py
martinfleis: Note that this may (shall) change in geopandas itself, allowing only `column` to be positional. But that is yet to be discussed.

lonboard/geopandas.py
martinfleis: > Another possibility is talking to geopandas about natively implementing this there.

We did and it is a hard no, sorry. That discussion was what kickstarted this PR in the first place.


> One idea is seeing about refactoring some of the upstream GeoPandas code, so that it isn't one big monolithic function. It would be nice if we could call geopandas.explore._explore_prep to manage a lot of the data prep that isn't related to the actual plotting in folium. Then we wouldn't have to duplicate all of the internals.

We can potentially talk about this but I would need to know which parts of the code you'd like to move out to a separate, presumable public, function.

knaaptime: the code here is already dramatically simpler than [explore](https://github.com/geopandas/geopandas/blob/main/geopandas/explore.py) which clocks in over 1k lines. That's mostly because this implementation uses mapclassify more efficiently and lonboard already has its own color application functions (and doesn't need to write to json, deal with branca, or generate a legend). Maybe some of the categorical or basemap handling would be useful once #494 is ready, but that would probably be the extent of it

lonboard/geopandas.py
kylebarron: I'd like this to be in line with future GeoPandas, so if GeoPandas is likely to add in `*`, then we might as well do it here

kylebarron:
I think it's probably worth having this in a `lonboard.geopandas` module. It can stay self contained and then we can document it could change in a backwards incompatible way across versions

A couple questions on the mechanics of how this works:

- Judging from [the pandas docs](https://pandas.pydata.org/docs/development/extending.html), as soon as `pd.api.extensions.register_dataframe_accessor` is called in global scope, the accessor is available, right? So with the current layout, once a user calls `import lonboard.geopandas`, the `GeoDataFrame.lb` accessor is already available?
- Should we have some API for the user to register/unregister the accessor?
- @martinfleis any preference on the namespacing? `.lb`? `.lon`? 
- Is there any way to support user-facing type hints here? I'm pretty sure the answer is usually no for these kind of extension mechanisms, but it would be nice if users could get some IDE help

lonboard/geopandas.py
martinfleis: It is not that easy as it is a breaking change... but it would be a sensible thing to do.

martinfleis:
> So with the current layout, once a user calls import lonboard.geopandas, the GeoDataFrame.lb accessor is already available?

yes

> Should we have some API for the user to register/unregister the accessor?

I don't think so.

> any preference on the namespacing? .lb? .lon?

no strong feelings, `lon` may conflict more with the real column name.

> Is there any way to support user-facing type hints here?

No clue.

lonboard/geopandas.py
kylebarron: Ok, let's keep the `*` here after `column`

kylebarron:
> > any preference on the namespacing? .lb? .lon?
> 
> no strong feelings, `lon` may conflict more with the real column name.
> 
> >

Good point. Let's use `lb`.

lonboard/geopandas.py
knaaptime: ok, I'll still update the arguments to put them in the same order (just because) and make sure the default values are consistent. My preference would be to stick lonboard-specific stuff like wireframe and elevation at the end of the signature (assuming they'll continue to pass through to the existing explore's **kwargs), though I can remove them entirely if that's still your preference?

kylebarron: > ok, I'll still update the arguments to put them in the same order (just because)

👍 that'll also make it easier to check what functionality is the same vs different from upstream

> My preference would be to stick lonboard-specific stuff like wireframe and elevation at the end of the signature

Let's remove any Lonboard-specific stuff for now and just get the minimal integration merged. That will be easier and faster to review and merge, and then we can add Lonboard-specific stuff as a follow up

knaaptime: ok i updated as discussed. Two small differences are that the `alpha` and `nan_color` arguments are at the end, after all the existing kwargs in the existing explore. In the folium-based version, transparency is handled py passing the opacity option to the style kwds, so alpha isn't available, and `nan_color` would be passed to 'color' in `missing_kwds`, which isn't provided here

I can drop those if we'd like. I dont have a good solution for transparency if we drop the alpha keyword. I could add a 'missing_kwds' argument that accepts color instead of exposing nan_color directly, which would mirror the folium behavior, though it would be kinda cumbersome to work with a dict jsut for that one option

knaaptime:
for some reason the precommit hook removes the `import lonboard.geopandas` line from the example notebook

kylebarron:
I think you can put `# noqa: F401` at the end of the import line

lonboard/geopandas.py
kylebarron: Even without a full-on xyzservices integration in `lonboard.basemap`, we can depend on xyzservices here and pass the URL on as a string

lonboard/geopandas.py
```diff
@@ -0,0 +1,318 @@
...
+                elevation: Series = gdf[elevation]
+            else:
+                raise ValueError(
+                    f"the designated height column {elevation} is not in the dataframe",
+                )
+            if not pd.api.types.is_numeric_dtype(elevation):
+                raise ValueError("elevation must be a numeric data type")
+        if elevation is not None:
+            layer_kwargs["extruded"] = True
+        """
```
kylebarron, line 244: is this a comment...? If we don't plan to use the code we should remove it

lonboard/geopandas.py
```diff
@@ -0,0 +1,318 @@
...
+
+        new_m: Map = viz(
```
kylebarron, line 246: this type hint can be inferred
```suggestion
        new_m = viz(
```

lonboard/geopandas.py
```diff
@@ -0,0 +1,318 @@
...
+    >>> xyz._query_name("carto db/positron")
+    >>> xyz._query_name("CARTO_DB_POSITRON")
+    >>> xyz._query_name("CartoDB.Positron")
+
+    """
+    name_clean = name.translate(_QUERY_NAME_TRANSLATION).lower()
+    if name_clean in _BASEMAP_PROVIDERS:
+        return _BASEMAP_PROVIDERS[name_clean]
+
+    raise ValueError(f"No matching provider found for the query '{name}'.")
```
kylebarron, line 318: we can remove this in favor of xyzservices

lonboard/geopandas.py
knaaptime: that sounds great, though i guess im not grokking how to do that, even with the recent changes to the basemap class. E.g. if i wanted to use darkmatter, I'd pass the url 'https://a.basemaps.cartocdn.com/dark_all/{z}/{x}/{y}.png', but I cant send that to the `basemap` keyword or to the MapLibre class. Aren't the providers in xyzservices exclusively raster tile servers (so not compatible with maplibre?)?

PR #872, opened 2025-09-16
fix: Fix rendering Arrow input with string view/binary view column

### LINKED ISSUE #871: Error loading child models or setting bounds
This seems to be triggered pretty easily when selecting bounds on the map, though I don't have an exact repro

kylebarron:
Closes #871. That's caused by a string view column that parquet-wasm is unable to parse. 

https://github.com/kylebarron/parquet-wasm/pull/786 fixes this issue. I should get a new stable release of parquet-wasm out soon, but for now we can use the beta to unblock.

This won't fail CI because the failure is on the JS side.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "lodash.throttle": "^4.1.1",
         "maplibre-gl": "^3.6.2",
         "memoize-one": "^6.0.0",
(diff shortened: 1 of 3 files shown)


kylebarron:
image
Now rendering correctly

PR #881, opened 2025-09-18
feat: New example with marimo and GeoDatafusion

kylebarron:


kylebarron:
https://github.com/user-attachments/assets/77f6a2b3-80c9-4524-8be2-79152746da1d

kylebarron:
I think that by keeping this video in this PR, we can link to it stably from the notebook 🤷‍♂️

PR #883, opened 2025-09-18
docs: Don't run mkdocs-jupyter on py files

kylebarron:
Turns out mkdocs-jupyter is clobbering the marimo example added in #881 

image


kylebarron:
Published new docs from this branch and it looks better

PR #884, opened 2025-09-21
fix DataFilterExtension get_filter_category

ATL2001:
as we discussed in discussion #711 I was unable to get the DataFilterExtension to work for filtering data with categories.

I finally spent some time looking into what was going on, and the `FilterValueAccessor` was being used for the get_filter_category property, but that accessor is set up to work in conjunction with the `filter_size` parameter of the DataFilterExtension.  I  created a new `FilterCategoryAccessor` accessor modeled after the `FilterValueAccessor` that works with the `category_size` parameter and does not do any casting of values to Float32, because looking at the DeckGL docs it appears that the category_filter should be able to work with string data.

Using the new accessor for the get_filter_category  on the data filter extension I'm now able to change `filter_categories` on the layer and the features are filtered as expected for numeric data.  What I've got still isn't working for filtering with string data, but I figured this was worth throwing out there to see if anyone else had an idea how to get the filter working with those as well.

I've added a notebook to the examples folder that I do not intend to actually check into the repo, which creates a DFE for a layer and links it to a widget to show that the category_filter is working for the numeric data, but not the string data.  If we get the strings working I'd be happy to make a better example notebook that showcases category filters

### DIFF
--- a/lonboard/layer_extension.py
+++ b/lonboard/layer_extension.py
@@ -5,6 +5,7 @@ import traitlets as t
 from lonboard._base import BaseExtension
 from lonboard.traits import (
     DashArrayAccessor,
+    FilterCategoryAccessor,
     FilterValueAccessor,
     FloatAccessor,
     PointAccessor,
@@ -353,10 +354,12 @@ class DataFilterExtension(BaseExtension):
         "filter_transform_size": t.Bool(default_value=True).tag(sync=True),
         "filter_transform_color": t.Bool(default_value=True).tag(sync=True),
         "get_filter_value": FilterValueAccessor(default_value=None, allow_none=True),
-        "get_filter_category": FilterValueAccessor(default_value=None, allow_none=True),
+        "get_filter_category": FilterCategoryAccessor(
+            default_value=None, allow_none=True,
+        ),
     }
 
-    filter_size = t.Int(None, min=1, max=4, allow_none=True).tag(sync=True)
+    filter_size = t.Int(None, min=0, max=4, allow_none=True).tag(sync=True)
     """The size of the filter (number of columns to filter by).
 
     The data filter can show/hide data based on 1-4 numeric properties of each object.
@@ -365,7 +368,7 @@ class DataFilterExtension(BaseExtension):
     - Default 1.
     """
 
-    category_size = t.Int(None, min=1, max=4, allow_none=True).tag(sync=True)
+    category_size = t.Int(None, min=0, max=4, allow_none=True).tag(sync=True)
     """The size of the category filter (number of columns to filter by).
 
     The category filter can show/hide data based on 1-4 properties of each object.
--- a/lonboard/traits.py
+++ b/lonboard/traits.py
@@ -823,6 +823,178 @@ class FilterValueAccessor(FixedErrorTraitType):
         return value.rechunk(max_chunksize=obj._rows_per_chunk)
 
 
+class FilterCategoryAccessor(FixedErrorTraitType):
+    """Validate input for `get_filter_category`.
+
+    A trait to validate input for the `get_filter_category` accessor added by the
+    [`DataFilterExtension`][lonboard.layer_extension.DataFilterExtension], which can
+    have between 1 and 4 values per row.
+
+
+    Various input is allowed:
+
+    - An `int` or `float`. This will be used as the value for all objects. The
+      `category_size` of the
+      [`DataFilterExtension`][lonboard.layer_extension.DataFilterExtension] instance
+      must be 1.
+    - A one-dimensional numpy `ndarray` with a numeric data type. Each value in the array will
+      be used as the value for the object at the same row index. The `category_size` of
+      the [`DataFilterExtension`][lonboard.layer_extension.DataFilterExtension] instance
+      must be 1.
+    - A two-dimensional numpy `ndarray` with a numeric data type. Each value in the array will
+      be used as the value for the object at the same row index. The `category_size` of
+      the [`DataFilterExtension`][lonboard.layer_extension.DataFilterExtension] instance
+      must match the size of the second dimension of the array.
+    - A pandas `Series` with a numeric data type. Each value in the array will be used as
+      the value for the object at the same row index. The `category_size` of the
+      [`DataFilterExtension`][lonboard.layer_extension.DataFilterExtension] instance
+      must be 1.
+    - A pyarrow [`FloatArray`][pyarrow.FloatArray], [`DoubleArray`][pyarrow.DoubleArray]
+      or [`ChunkedArray`][pyarrow.ChunkedArray] containing either a `FloatArray` or
+      `DoubleArray`. Each value in the array will be used as the value for the object at
+      the same row index. The `category_size` of the
+      [`DataFilterExtension`][lonboard.layer_extension.DataFilterExtension] instance
+      must be 1.
+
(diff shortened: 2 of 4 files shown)


ATL2001:
a couple screenshots to show the selector changing the filtered points on my map:
image
image

lonboard/layer_extension.py
```diff
@@ -353,10 +354,13 @@ class DataFilterExtension(BaseExtension):
...
         "get_filter_value": FilterValueAccessor(default_value=None, allow_none=True),
-        "get_filter_category": FilterValueAccessor(default_value=None, allow_none=True),
+        "get_filter_category": FilterCategoryAccessor(
+            default_value=None,
+            allow_none=True,
+        ),
     }
 
-    filter_size = t.Int(None, min=1, max=4, allow_none=True).tag(sync=True)
+    filter_size = t.Int(None, min=0, max=4, allow_none=True).tag(sync=True)
```
kylebarron: The JS [docs say](https://deck.gl/docs/api-reference/extensions/data-filter-extension)

> Set to `0` to disable numeric filtering

I think instead of exposing that same behavior to Python, we should override `None` to mean "disabled". So when the Python side passes in `None` it translates to passing `0` into JS.

Similarly, we should change the default value here to `1` to match JS, now that `None` doesn't mean "undefined" but rather "defined and null"

lonboard/layer_extension.py
```diff
@@ -365,7 +369,7 @@ class DataFilterExtension(BaseExtension):
     - Default 1.
     """
 
-    category_size = t.Int(None, min=1, max=4, allow_none=True).tag(sync=True)
+    category_size = t.Int(None, min=0, max=4, allow_none=True).tag(sync=True)
```
kylebarron: Ditto for the same behavior as above, though the default here can be `None`

lonboard/traits.py
```diff
@@ -823,6 +823,178 @@ def validate(
         return value.rechunk(max_chunksize=obj._rows_per_chunk)
 
 
+class FilterCategoryAccessor(FixedErrorTraitType):
```
kylebarron: Can you add some tests for this? There are some example tests in `test_traits.py` and you can look at https://github.com/developmentseed/lonboard/pull/917 for more examples.

It might be worth making `test_traits` into a folder and having a file specifically for `test_traits/test_filter_extension.py`

src/model/extension.ts
```diff
@@ -146,19 +146,25 @@ export class DataFilterExtension extends BaseExtensionModel {
   }
 
   extensionInstance(): _DataFilterExtension | null {
-    if (isDefined(this.filterSize)) {
+    if (isDefined(this.filterSize) && isDefined(this.categorySize)) {
```
kylebarron: Can we avoid having this extra case? Can we leave it to the user to define their filter sizes and category sizes correctly? I.e. why would this case need to be different from the case below it?

ATL2001: Oh, whoops! If I change that `&&` to an `||` then we won't need either of the the other `else if` blocks at all since in the creation of props it's checking `isDefined` too

lonboard/traits.py
ATL2001: ran out of time tonight, but I'll see what I can do in the next couple evenings, or maybe this weekend

lonboard/layer_extension.py
kylebarron: `int` or `None`. You can set `None` to turn off the numeric filtering, e.g. to use only category filtering, is that right?

ATL2001:
ugh, I don't know what I've done over here (I honestly don't think I did anything, but clearly I must have). Friday night I went in to modify that docstring and start making some tests, and I started getting the `GL_INVALID_OPERATION: Vertex shader input type does not match the type of the bound vertex attribute` error that was happening when I was sending a string column to the category filter, but now it's showing up for the int and float columns too. 😭 

I'm starting to think it may be a better use of my time to abandon this branch and start over from scratch 😞

kylebarron:
Do you know if this branch works again with the latest fix on `main`?

ATL2001:
yes, I just pulled the updates in, and it is functional again 🎉 .  I'll try to carve out some time in the evenings this week to see if I can wire up some sort of tests.

ATL2001:
@kylebarron while I'm updating the data filter extension example to add a section for categorical filtering, do you mind if I change the existing DataFilterExtension example notebook?  

What I'm thinking of doing is making a geopandas dataframe of 1 million random points with columns for x/y/a random int(1-5) and then hooking a MultiRangeSlider to these columns, so the person playing with the sliders won't need to know anything about the data being displayed to understand how changing the sliders limits the data displayed.

image

kylebarron:
> @kylebarron while I'm updating the data filter extension example to add a section for categorical filtering, do you mind if I change the existing DataFilterExtension example notebook?

Could you add another new notebook specifically for category filtering?
 
> What I'm thinking of doing is making a geopandas dataframe of 1 million random points with columns for x/y/a random int(1-5) and then hooking a MultiRangeSlider to these columns, so the person playing with the sliders won't need to know anything about the data being displayed to understand how changing the sliders limits the data displayed.

I don't think it's that hard to understand that the sliders change the numeric values in the frontend, even if in the case of the existing notebook it's download speeds? The names of the sliders should be pretty self explanatory

src/model/extension.ts
```diff
@@ -142,19 +142,15 @@ export class DataFilterExtension extends BaseExtensionModel {
...
-      const props = {
-        ...(isDefined(this.filterSize) ? { filterSize: this.filterSize } : {}),
-      };
-      // console.log("ext props", props);
-      return new _DataFilterExtension(props);
-    } else if (isDefined(this.categorySize)) {
+    if (isDefined(this.filterSize) || isDefined(this.categorySize)) {
       const props = {
+        ...(isDefined(this.filterSize)
+          ? { filterSize: this.filterSize != null ? this.filterSize : 0 }
```
kylebarron: In JS you always need to use `!==` instead of `!=`
```suggestion
          ? { filterSize: this.filterSize !== null ? this.filterSize : 0 }
```

src/model/extension.ts
```diff
@@ -142,19 +142,15 @@ export class DataFilterExtension extends BaseExtensionModel {
...
+          : {}),
         ...(isDefined(this.categorySize)
-          ? { categorySize: this.categorySize }
+          ? { categorySize: this.categorySize != null ? this.categorySize : 0 }
```
kylebarron:
```suggestion
          ? { categorySize: this.categorySize !== null ? this.categorySize : 0 }
```

kylebarron:
Ah I missed the last few copilot suggestions 🥲

mkdocs.yml
```diff
@@ -47,6 +47,7 @@ nav:
       - examples/kontur_pop.ipynb
       - examples/migration.ipynb
       - examples/data-filter-extension.ipynb
+      - examples/data-filter-extension-categorical.ipynb
```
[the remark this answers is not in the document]
kylebarron: I think it's fine to add it to the docs as long as you're ok with it @ATL2001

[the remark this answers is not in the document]
ATL2001: I think copilot was thinking this was the original throw away notebook I had in here, but this one is intentional for showcasing the usage of the DFE for categoricals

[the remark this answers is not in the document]
kylebarron: The notebook example looks great!

Issue #842, opened 2025-08-06
Fullscreen


wildintellect:
**Is your feature request related to a problem? Please describe.**
It would be great to have the ability to fullscreen the map with a button to toggle.

**Describe the solution you'd like**
Folium has a great example to mimic https://python-visualization.github.io/folium/latest/user_guide/plugins/fullscreen.html

**Describe alternatives you've considered**

Sidecar works in a similar way if you pop it to a tab, but that requires an external extension configured correctly.

kylebarron:
I don't know how to implement this but a PR would be welcome!

wildintellect:
Neither do I, but we'll find someone who does. Looks like Folium leverages existing js code in Leaflet https://github.com/python-visualization/folium/blob/3fb4422439d926cd6ed0a644da0b15fc740e69f3/folium/plugins/fullscreen.py#L8
...
https://github.com/Leaflet/Leaflet.fullscreen

kylebarron:
Looks like deck.gl actually has something similar: https://deck.gl/docs/api-reference/widgets/fullscreen-widget FullscreenWidget | deck.gl

kylebarron:
Did a little exploration in https://github.com/developmentseed/lonboard/pull/848. Maybe that'll be a useful starting point for a frontend person.

PR #890, opened 2025-09-24
feat: Switch to using MapboxOverlay

kylebarron:
This revisits and supersedes https://github.com/developmentseed/lonboard/pull/437. 

### Change list

- Add `@deck.gl/mapbox` dependency
- Add `FullscreenControl` and `NavigationControl`. These are placed on the top-left of the map for now, because the bounding box selection is already on the top right. But perhaps we should move the bbox selection to the left side? 🤷‍♂️ 

### Questions

- Does it make sense to use controls from _Maplibre_ instead of the related "widgets" from Deck.gl v9? 

See also more background information in https://github.com/developmentseed/lonboard/pull/437:

> There's a wider question here: should Maplibre be a child of Deck or should Deck be a child of Maplibre? The Deck docs list [some limitations](https://deck.gl/docs/api-reference/mapbox/overview#limitations):
> 
> > * When using deck.gl's multi-view system, only one of the views can match the base map and receive interaction. See [using MapboxOverlay with multi-views](./mapbox-overlay.md#multi-view-usage) for details.
> > * When using deck.gl as Mapbox layers or controls, `Deck` only receives a subset of user inputs delegated by `Map`. Therefore, certain interactive callbacks like `onDrag`, `onInteractionStateChange` are not available.
> > * Mapbox/Maplibre's terrain features are partially supported. When a terrain is used, the camera of deck.gl and the base map should synchronize, however the deck.gl data with z=0 are rendered at the sea level and not aligned with the terrain surface.
> > * Only Mercator projection is supported. Mapbox adaptive projection is not supported as their API doesn't expose the projection used.
> > * The `position` property in `viewState` has no equivalent in mapbox-gl.
> 
> #### Maplibre as a child of Deck (not using MapboxOverlay)
> * We should be able to have more low level control over visualization and be less tied to maplibre.
> * For example, it would be really cool to visualize a side-by-side map, where you could swipe to see a before-and-after with two datasets. This option is required for this feature. Otherwise, when using `MapboxOverlay` [you can't have multiple interactive deck views](https://deck.gl/docs/api-reference/mapbox/mapbox-overlay#multi-view-usage).
> * This also would keep us less tied to geospatial. So in theory non-geospatial users could use lonboard without a basemap, and still get to use all of lonboard's utilities for data serialization.
> 
> #### Deck as a child of Maplibre (using MapboxOverlay)
> * Can use all Maplibre and Maplibre-compatible controls. In theory this should even include things like [mapbox-gl-draw](https://github.com/mapbox/mapbox-gl-draw). ([Upstream example](https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-draw/)). In theory we wouldn't have to write our own, like we're doing in [Select by bounding box #417](https://github.com/developmentseed/lonboard/pull/417)
> * Note that `fly_to` is now broken. This is because deck is no longer maintaining the view state itself.

I think eventually it may make sense to allow rendering deck.gl without any sort of geospatial basemap, but that's a future topic.

Closes https://github.com/developmentseed/lonboard/issues/436, closes https://github.com/developmentseed/lonboard/issues/842

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -2,7 +2,12 @@ import * as React from "react";
 import { useEffect, useCallback, useState } from "react";
 import { createRender, useModelState, useModel } from "@anywidget/react";
 import type { Initialize, Render } from "@anywidget/types";
-import Map, { useControl } from "react-map-gl/maplibre";
+import Map, {
+  useControl,
+  FullscreenControl,
+  NavigationControl,
+  ScaleControl,
+} from "react-map-gl/maplibre";
 import { MapViewState, PickingInfo, type Layer } from "@deck.gl/core";
 import { BaseLayerModel, initializeLayer } from "./model/index.js";
 import type { WidgetModel } from "@jupyter-widgets/base";
@@ -249,7 +254,6 @@ function App() {
         <div className="bg-red-800 h-full w-full relative">
           <Map
             reuseMaps
-            // projection="globe"
             id={`map-${mapId}`}
             initialViewState={
               ["longitude", "latitude", "zoom"].every((key) =>
@@ -258,19 +262,14 @@ function App() {
                 ? initialViewState
                 : DEFAULT_INITIAL_VIEW_STATE
             }
-            // boxZoom={false}
-            // dragRotate={false}
-            // maxPitch={0}
             mapStyle={mapStyle || DEFAULT_MAP_STYLE}
-            attributionControl={{ customAttribution }}
-            onContextMenu={(e) => {
-              console.log("hi");
-
-              e.originalEvent.stopPropagation();
-              // e.stopPropagation();
-              e.preventDefault();
+            attributionControl={{
+              customAttribution,
             }}
           >
+            <FullscreenControl position="top-left" />
+            <NavigationControl position="top-left" />
+            <ScaleControl position="bottom-left" />
             <DeckGLOverlay
               // interleaved
               style={{ width: "100%", height: "100%" }}
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -2,8 +2,7 @@ import * as React from "react";
 import { useEffect, useCallback, useState } from "react";
 import { createRender, useModelState, useModel } from "@anywidget/react";
 import type { Initialize, Render } from "@anywidget/types";
-import Map from "react-map-gl/maplibre";
-import DeckGL from "@deck.gl/react";
+import Map, { useControl } from "react-map-gl/maplibre";
 import { MapViewState, PickingInfo, type Layer } from "@deck.gl/core";
 import { BaseLayerModel, initializeLayer } from "./model/index.js";
 import type { WidgetModel } from "@jupyter-widgets/base";
@@ -13,6 +12,10 @@ import { v4 as uuidv4 } from "uuid";
 import { Message } from "./types.js";
 import { flyTo } from "./actions/fly-to.js";
 import { useViewStateDebounced } from "./state";
+import {
+  MapboxOverlay as DeckOverlay,
+  MapboxOverlayProps,
+} from "@deck.gl/mapbox";
 
 import { MachineContext, MachineProvider } from "./xstate";
 import * as selectors from "./xstate/selectors";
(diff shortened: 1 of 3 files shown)


src/index.tsx
```diff
@@ -38,6 +46,13 @@ const DEFAULT_INITIAL_VIEW_STATE = {
 const DEFAULT_MAP_STYLE =
   "https://basemaps.cartocdn.com/gl/positron-nolabels-gl-style/style.json";
 
+function DeckGLOverlay(props: MapboxOverlayProps) {
+  const overlay = useControl(() => new DeckOverlay(props));
+
+  overlay.setProps(props);
+  return null;
+}
```
kylebarron, line 54: Is this the standard way to use `DeckOverlay` in react? @felixpalmer

src/index.tsx
```diff
@@ -237,57 +252,65 @@ function App() {
...
+                    viewState;
+                  setViewState({
+                    longitude,
+                    latitude,
+                    zoom,
+                    pitch,
+                    bearing,
+                  });
+                }
+              }}
```
kylebarron, line 310: todo: I need to check that deck.gl still sends these events even when the view state is managed by maplibre. 

For background: This sets the current view state on the _model_ so that it can be updated on the Python side, which can be accessed via `Map.view_state`.

src/index.tsx
felixpalmer: Yes https://deck.gl/docs/api-reference/mapbox/mapbox-overlay

felixpalmer:
> Does it make sense to use controls from Maplibre instead of the related "widgets" from Deck.gl v9?

Up to you, but worth noting many more widgets are coming in 9.2 https://github.com/visgl/deck.gl/pull/9784

kylebarron:
Is it possible to use the deck.gl widgets even with MapboxOverlay?

Also, with MapboxOverlay it's not possible to use two views, right? Or at least they can't both have a maplibre map?

kylebarron:
From @felixpalmer, suggestion on how to switch ctrl and shift:
```diff
diff --git a/examples/get-started/react/maplibre/app.jsx b/examples/get-started/react/maplibre/app.jsx
index 238df7a16..c08899c13 100644
--- a/examples/get-started/react/maplibre/app.jsx
+++ b/examples/get-started/react/maplibre/app.jsx
@@ -22,6 +22,14 @@ const INITIAL_VIEW_STATE = {
 };
 
 const MAP_STYLE = 'https://basemaps.cartocdn.com/gl/positron-gl-style/style.json';
+
+function patchDragMoveHandler(handler) {
+  const originalCorrectEvent = handler._moveStateManager._correctEvent;
+  handler._moveStateManager._correctEvent = e => {
+    return originalCorrectEvent({button: e.button, ctrlKey: e.shiftKey});
+  };
+}
+
 function DeckGLOverlay(props) {
   const overlay = useControl(() => new DeckOverlay(props));
   overlay.setProps(props);
@@ -31,6 +39,12 @@ function DeckGLOverlay(props) {
 function Root() {
   const [selected, setSelected] = useState(null);
 
+  const handleMapLoad = (event) => {
+    const map = event.target;
+    patchDragMoveHandler(map.dragRotate._mousePitch);
+    patchDragMoveHandler(map.dragRotate._mouseRotate);
+  };
+
   const layers = [
     new GeoJsonLayer({
       id: 'airports',
@@ -61,7 +75,7 @@ function Root() {
   ];
 
   return (
-    <Map initialViewState={INITIAL_VIEW_STATE} mapStyle={MAP_STYLE}>
+    <Map initialViewState={INITIAL_VIEW_STATE} mapStyle={MAP_STYLE} onLoad={handleMapLoad} boxZoom={false}>
       {selected && (
         <Popup
           key={selected.properties.name}
```

felixpalmer:
@HarelM is there any better way to switch the SHIFT/CTRL keys?

felixpalmer:
@kylebarron regarding multiview: https://deck.gl/docs/api-reference/mapbox/mapbox-overlay#multi-view-usage, correct only one can have the maplibre map.

It turns out you can add the deck widgets, just not the ones used to control the view state. Not sure how well supported this is though

```jsx
   return (
    <Map initialViewState={INITIAL_VIEW_STATE} mapStyle={MAP_STYLE} onLoad={handleMapLoad} boxZoom={false}>
       {selected && (
         <Popup
           key={selected.properties.name}
@@ -73,7 +101,10 @@ function Root() {
           {selected.properties.name} ({selected.properties.abbrev})
         </Popup>
       )}
      <DeckGLOverlay layers={layers} /* interleaved*/ widgets={[
        new FullscreenWidget({placement: 'bottom-left'})
      ]
      }/>
       <NavigationControl position="top-left" />
     </Map>
   );

```

HarelM:
> @HarelM is there any better way to switch the SHIFT/CTRL keys?

Not at the moment, but I think there was a PR that requested it, but didn't make it to the finish line. I think that configurable preset might be needed as configuring for each handler can lead to "keys configuration collision".

Issue #886, opened 2025-09-22
Support Maplibre GlobeView


felixpalmer:
**Is your feature request related to a problem? Please describe.**
deck.gl 9.1 added integration with Maplibre's GlobeView: https://deck.gl/docs/whats-new#globe-view--maplibre

**Describe the solution you'd like**
Lonboard should allow the user to switch to globe view

**Additional context**
https://deck.gl/examples/maplibre

<img width="800" height="800" alt="Image" src="https://github.com/user-attachments/assets/62cdcc12-38a1-481d-9717-02d65652aa23" />

kylebarron:
I hadn't realized that deck.gl now supports GlobeView. I think it would be great to have this supported.

I can prototype the Python API if you want a start of how we would integrate globe view. Would that help?

felixpalmer:
>  can prototype the Python API 

Sounds great, I imagine it will just be a config option somewhere to set `projection = 'mercator' | 'globe'`

Can you point me at where the Maplibre integration code lives? I can take a look at hacking in globe view as a PoC

kylebarron:
I think all Map integration lives in this file: 
https://github.com/developmentseed/lonboard/blob/9632fe1416a63ccb61c2c1671042f51adcc1ba20/src/index.tsx

vgeorge:
I did a quick experiment on setting `map.setProjection({ type: "globe" });`:

![Image](https://github.com/user-attachments/assets/6800d137-4740-4082-bd9c-f3907cebd81a)

The globe basemap renders correctly, but the deck.gl layer floats above the surface since, in interleaved mode, deck.gl doesn’t inherit MapLibre’s projection and keeps rendering in flat space. The MapboxOverlay work in #890 should help get the globe view right.

kylebarron:
Yes, I think we need to use `MapboxOverlay`, but I also want to support rendering deck.gl without Maplibre for use cases that don't want to /can't be tied to Maplibre. So I think we'll have to split into two underlying rendering modes. 

It's just a bit annoying that (at least by default) the interaction controls between a deck.gl-controlled and MapLibre-controlled map are different https://github.com/developmentseed/lonboard/pull/890#issuecomment-3332436698

PR #906, opened 2025-10-07
Feat: add Globe View and automated UI testing with Playwright

vgeorge:
Contributes to #886

Related work

* [[#890](https://github.com/developmentseed/lonboard/pull/890)](https://github.com/developmentseed/lonboard/pull/890)
* [[#437](https://github.com/developmentseed/lonboard/pull/437)](https://github.com/developmentseed/lonboard/pull/437)

### Changes introduced

* Bump Deck.gl to `v9.1.15`
* Add Playwright setup for automated UI testing
* Introduce a Mapbox overlay–based Globe View (`GlobeMap`), rendered when `projection="globe"` is set

### Description

This PR takes a first, minimal step toward supporting the Globe View by introducing the Mapbox overlay approach.

Playwright testing is added to validate key UI interactions and prevent regressions. The current setup includes a test that verifies bounding box selection on the flat map. We can expand it later to cover other existing features once the testing framework stabilizes.

### How to review

* Run `npm run test:e2e`, all tests should pass
* Open any existing example and set `projection="globe"`, the map should render data using the Globe View (bbox selection is not yet supported)

### Preview

globe

### Follow-ups

* Confirm if the approach is viable for the Globe View
* Add Playwright tests to CI

@kylebarron this is ready for a first review.


kylebarron:
Thanks for this! I'll write up some more thoughts, but first, could we split the playwright tests into a separate PR? I think that part is quite valuable on its own and will be relatively easy to merge while the globe view might take a little longer to architect.

tests/e2e/helpers/deckgl/types.ts
kylebarron, line 1: Are these types not exported from a deck.gl module?

vgeorge:
@kylebarron sounds good, I'll create a separated PR just for the playwright implementation.

tests/e2e/helpers/deckgl/types.ts
vgeorge: They probably are, I'll check.

kylebarron:
> deck.gl now works seamlessly with the MapLibre v5 [globe view](https://maplibre.org/maplibre-gl-js/docs/examples/globe-vector-tiles/) for all three [Basemap Intergration Modes](https://deck.gl/docs/developer-guide/base-maps/using-with-maplibre#integration-modes)

It's possible we may not need to use MapboxOverlay at all for globe view https://deck.gl/docs/whats-new#globe-view--maplibre

kylebarron:
See https://github.com/developmentseed/lonboard/pull/908 for what I'm thinking of for the Python side, and then when that's serialized to the JS side, we'd switch on the `basemap.mode` attribute to decide whether to render MapboxOverlay or not

kylebarron:
The failing test is just a formatting lint to use double string quotes in your ipynb

src/index.tsx
```diff
@@ -94,6 +95,14 @@ function App() {
 
   const [justClicked, setJustClicked] = useState<boolean>(false);
 
+  const deckRef = useRef<DeckGLRef | null>(null);
```
kylebarron, line 99: Can you add a comment here explaining what this does?

tests/e2e/fixtures/.gitignore
```diff
@@ -0,0 +1 @@
+.ipynb_checkpoints/
```
kylebarron: You can just add this in the top-level `.gitignore`; I think that'll be simpler

vgeorge:
@kylebarron I’ve updated the PR title and description, and addressed your comments. Some type definitions aren’t exposed by Deck.gl. I didn’t have time to look deeper, so I’d suggest we revisit that later. We should also document how to install the pre-commit checks for Python. Regarding the CI checks, that’s definitely a must but not sure about complexity. Let’s track it in a separate ticket.

Issue #375, opened 2024-02-20
Add support for OrthographicView


cornhundred:
## Context
I would like to visualize non geographic data (e.g., spatial multi-omic biological data) using lonboard. I belive that we would have the add the option for the [OrthographicView](https://deck.gl/docs/api-reference/core/orthographic-view) class in deck.gl.

## Issue
I think that we just need to expose the view classes to lonboard to allow users to use the OrthograhicView.

## Acceptance-Criteria
I'm not sure what these would be.

kylebarron:
This is related to https://github.com/developmentseed/lonboard/issues/322.

I think that `View` is always passed in to `Deck` directly, not the `Layer` class, right? In that case it should be relatively simple.

As a general statement, much of lonboard so far has geospatial assumptions baked in, but I think nothing should preclude using non-geospatial data

cornhundred:
Thanks, that is my understanding too that it is passed into Deck directly. 

Here's how I'm setting the view on an ObservableHQ [notebook](https://observablehq.com/@cornhundred/mouse-brain-experiment)

```
const view = new deck.OrthographicView({id: 'ortho'})

deck.DeckGL({
    container,
    views:[view],
    initialViewState:INITIAL_VIEW_STATE,
    controller: true,
  });
```

I can have a look at lonboard to see if I can figure out how to make a pull request.

kylebarron:
A PR would be great. In general, models are defined first in TS https://github.com/developmentseed/lonboard/blob/main/src/model/layer.ts (maybe create a new file for `view.ts`?), then create wrapper classes (subclassing from `ipywidgets.Widget`) that map to the deck.gl api: https://github.com/developmentseed/lonboard/blob/4daca9bb36013a56a998dc1507d03058f582b141/lonboard/_layer.py#L288-L357

Issue #905, opened 2025-10-06
composable maps based on `ipywidgets`


keewis:
In xarray-contrib/xdggs#183 I'm trying to find a way to compose maps, either by taking all layers and adding them to a new map (this does not yet take data transfer into account), or by arranging multiple maps into a `ipywidgets.GridBox`.

This works pretty well already, but I'm wondering if it would make sense to upstream this into `lonboard`? If not, I'd be willing to open a PR to add this to the "ecosystem" page (if that's the right place).

kylebarron:
What are you trying to achieve here? Can you expand on what you mean by "compose maps"? Do you just mean a side-by-side view?

keewis:
side-by-side is the second use-case (which I'm implementing using `GridBox`) but the most important one is combining two maps such that the layers are displayed in the same map.

Ideally, I'd have a separate container that delays data transfers until the map is displayed (so that data transfer is lazy), but I have not yet figured out how the protocol used by `ipywidgets` / `anywidget` to interact with `jupyterlab` works.

kylebarron:
> side-by-side is the second use-case

I think side-by-side might become in scope for handling on the JS side natively

> the most important one is combining two maps such that the layers are displayed in the same map

Sorry I still don't understand. How is this different from `Map([*existing_map1.layers, *existing_map2.layers])`?

keewis:
I have a wrapper in `xdggs` that returns one `Map` instance per dataset containing one or more layers. If I have multiple datasets, I get multiple `Map` instances that I may want to overlay (which happens by adding all layers from the other maps to the first):
```python
map1 = ds1["var1"].dggs.explore()
map2 = ds2["var2"].dggs.explore()
# overlay map1 with layers from map2
```

This may be the due to the abstraction of returning map objects, and so an API like:
```python
map = lonboard.Map()
ds1["var1"].dggs.explore(map=map)  # calls map.add_layer
ds2["var2"].dggs.explore(map=map)
```
may be better, but what I came up with is `hvplot`-style combinations:
```python
ds1["var1"].dggs.explore() & ds2["var2"].dggs.explore()
```

What I was wondering is whether this is something that could live within `lonboard.Map` (whatever the API), or whether I should continue using a wrapper object.

keewis:
> I think side-by-side might become in scope for handling on the JS side natively

[...]

Issue #718, opened 2024-12-15
BBOX for Map


leonard-psu:
**Is your feature request related to a problem? Please describe.**
Using side by side maps that are in sync.  Map #1 uses a small geoparquet file, the other Map #2 uses a very large geoparquet file. When the user interacts with Map #1 I want to use the map bounding-box/view-extent to query the data to update Map #2.

**Describe the solution you'd like**
Public function to get the bounding box of the Map. i.e. map.getBounds()
Event state change to Map's bounding box.

**Describe alternatives you've considered**
Tested access to compute_view (private fn, i.e. wont work) to access bbox from pydeck.ViewPort.

**Additional context**

Hoping to be pointed in the right direction for a friendly way to get bbox.

kylebarron:
There's an example in the docs for linked maps: https://developmentseed.org/lonboard/latest/examples/linked-maps/

Issue #658, opened 2024-09-30
Enable repeating map view


kylebarron:
https://deck.gl/docs/whats-new#world-repeating-in-web-mercator-maps

kylebarron:
Actually it turns out this is now enabled by default on the deck.gl side; I don't have to explicitly enable it

<img width="958" height="746" alt="Image" src="https://github.com/user-attachments/assets/7840216c-428e-48c2-a04b-72ec1c088302" />

PR #908, opened 2025-10-08
feat: Support globe view

kylebarron:
### Change list

- Ensure that the default value of `Map.basemap` is `MaplibreBasemap` if no value of `basemap` was passed. Add a test for this.
- Define various View classes to map to deck.gl views.
- Define View models on the TS side.
- When a GlobeView is passed, pass `projection="globe"` to Maplibre
- Render a dark background on the canvas when in globe view 
- Add `views` parameter to `Map`. For now this only supports a single view instance, but hopefully in the future it'll support more.

This PR also sets the stage for non-geospatial map rendering and for multi-view support.

### todo:

- [ ] Should we name the Map attribute `view` instead of `views`? Hopefully soon we'll support multi-views, though maybe not with a maplibre basemap. (#965 )
- [ ] View state validation (https://github.com/developmentseed/lonboard/issues/948)
- [ ] add something like `projection="globe"` to `viz` (https://github.com/developmentseed/lonboard/issues/949)

With this PR, to generate a map with globe view, someone would call

```py
from lonboard import Map
from lonboard.view import GlobeView

lonboard.Map(
	layer,
	views=GlobeView(),
	basemap=MaplibreBasemap(mode="interleaved"),
)
```

## Globe view:

image

Closes https://github.com/developmentseed/lonboard/issues/886 (globe view), closes https://github.com/developmentseed/lonboard/issues/375 (orthographic view) (though might need more work to allow non-spatial data)

Relates to 

- https://github.com/developmentseed/lonboard/issues/905 (multi-views), 
- https://github.com/developmentseed/lonboard/issues/718 (multi-view), 
- https://github.com/developmentseed/lonboard/issues/658 (repeating map view), 
- https://github.com/developmentseed/lonboard/issues/637 (multiple linked plots),
- https://github.com/developmentseed/lonboard/issues/494 (basemap refactor), 
- https://github.com/developmentseed/lonboard/issues/404 (swipe between two maps),
- https://github.com/developmentseed/lonboard/issues/375 (orthographic view)


lonboard/_html_export.py
kylebarron, line 1: This is a small separate refactor to make it easier to read `_map.py` by removing the HTML export functionality

lonboard/_map.py
```diff
@@ -151,6 +131,35 @@ def on_click(self, callback: Callable, *, remove: bool = False) -> None:
     _esm = bundler_output_dir / "index.js"
     _css = bundler_output_dir / "index.css"
 
+    _has_click_handlers = t.Bool(default_value=False, allow_none=False).tag(sync=True)
```
kylebarron, line 181: This diff is from moving `view_state` lower so that it's immediately after the new `views` attribute; nothing was changed except for adding `views`

lonboard/_map.py
```diff
@@ -219,15 +211,15 @@ def on_click(self, callback: Callable, *, remove: bool = False) -> None:
     - Default: `5`
     """
 
-    basemap_style = BasemapUrl(CartoBasemap.PositronNoLabels)
-    """
-    A URL to a MapLibre-compatible basemap style.
+    basemap = t.Instance(MaplibreBasemap, allow_none=True).tag(
```
kylebarron: This allows users to customize the basemap, and to turn off the basemap altogether by passing `None`.

Before merging we may keep `basemap_style` around to not break backwards compatibility, but I think documenting that the new API is
```py
Map(basemap=MaplibreBasemap(basemap_style=basemap_style))
```
isn't too bad

lonboard/basemap.py
```diff
@@ -43,3 +49,51 @@ class CartoBasemap(str, Enum):
...
+
+    def __init__(
+        self,
+        *,
+        mode: Literal[
+            "interleaved",
+            "overlaid",
+            "reverse-controlled",
+        ] = "reverse-controlled",
+        basemap_style: str | CartoBasemap,
```
kylebarron: Here the user can explicitly choose a rendering mode between `interleaved` (mapbox overlay set to interleaved), `overlaid` (mapbox overlay not using interleaved), or `reverse-controlled`, the existing behavior, which doesn't use mapbox overlay

lonboard/view.py
kylebarron, line 1: This file implements Python models for deck.gl [Views](https://deck.gl/docs/developer-guide/views), allowing a definition of a globe view and of non-geospatial views.

src/model/view.ts
kylebarron, line 1: This defines the TS/JS side of the view models, with the ability to create deck `View` objects

vgeorge:
@kylebarron The general approach seems quite comprehensive. I’d suggest implementing the different modes more progressively, as this will likely require some refactoring on the JS side. In hindsight, adding XState wasn’t the best choice on my part, it makes state management a bit harder to evolve. I’ll open a ticket to propose an alternative. I think we can still address this technical debt while it’s not too complex, but not on this PR.

kylebarron:
I _think_ despite what are a bunch of changes here on the Python side, the JS side only needs to change to support two modes: One with mapbox overlay and one without. 

We can also merge this PR or something similar without exposing the classes to users, so we can keep developing on the JS side

vgeorge:
@kylebarron sounds good. Now that we merged the Playwright PR we can add specs to ensure JS features continue to work as expected.

lonboard/_html_export.py
kylebarron: Moved to https://github.com/developmentseed/lonboard/pull/934

kylebarron:
The basemap refactor was extracted to https://github.com/developmentseed/lonboard/pull/935

src/model/base.ts
```diff
@@ -17,10 +17,6 @@ export abstract class BaseModel {
     this.callbacks.set("change", updateStateCallback);
   }
 
-  async loadSubModels() {
-    return;
-  }
```
kylebarron, line 22: This was unused.

PR #921, opened 2025-10-14
feat: Support two render modes: Standard/deck.gl-first and MapboxOverlay

kylebarron:
As described in the [upstream deck.gl docs](https://deck.gl/docs/developer-guide/base-maps/using-with-maplibre), there are three ways to support using deck.gl with Maplibre: _interleaved_, _overlaid_, and _reverse-controlled_. The first two are supported via `MapboxOverlay` with a prop `interleaved: true|false`, while the latter is implemented by having Maplibre be a child of the deck.gl Map.

There are worthwhile reasons to support all of these modes. We need to use interleaved or overlaid to support globe view, while reverse-controlled better supports multiple deck.gl views.

This PR refactors the map component in `index.tsx` into two separate React components: a deck.gl-first renderer (i.e. "reverse-controlled") and a MapboxOverlay renderer.

The idea is that this will pair with #908 to give users more control over various ways of rendering maps.

This is backwards-compatible because we default to reverse-controlled, the existing default.

Closes https://github.com/developmentseed/lonboard/pull/890, closes https://github.com/developmentseed/lonboard/pull/437, for https://github.com/developmentseed/lonboard/issues/886, for https://github.com/developmentseed/lonboard/issues/718

### DIFF
--- a/src/renderers/overlay.tsx
+++ b/src/renderers/overlay.tsx
@@ -25,13 +25,8 @@ function DeckGLOverlay(props: MapboxOverlayProps) {
  */
 const OverlayRenderer: React.FC<MapRendererProps> = (mapProps) => {
   // Remove maplibre-specific props before passing to DeckGL
-  const {
-    mapStyle,
-    customAttribution,
-    initialViewState,
-    // deckRef,
-    ...deckProps
-  } = mapProps;
+  const { mapStyle, customAttribution, initialViewState, ...deckProps } =
+    mapProps;
   return (
     <Map
       reuseMaps
@@ -41,7 +36,6 @@ const OverlayRenderer: React.FC<MapRendererProps> = (mapProps) => {
       style={{ width: "100%", height: "100%" }}
     >
       <DeckGLOverlay
-        // ref={deckRef}
         // https://deck.gl/docs/api-reference/core/deck#_typedarraymanagerprops
         _typedArrayManagerProps={{
           overAlloc: 1,


lonboard/_map.py
```diff
@@ -179,6 +179,8 @@ def on_click(self, callback: Callable, *, remove: bool = False) -> None:
     Indicates if a click handler has been registered.
     """
 
+    render_mode = t.Unicode(default_value="deck-first").tag(sync=True)
```
kylebarron, line 182: This will be further improved on from the Python side to connect with #908

src/renderers/overlay.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * MapboxOverlay. This approach gives the base map more control and can
+ * enable features like interleaved rendering between map and deck layers.
+ */
+const OverlayRenderer: React.FC<MapRendererProps> = (mapProps) => {
+  // Remove maplibre-specific props before passing to DeckGL
+  const {
+    mapStyle,
+    customAttribution,
+    initialViewState,
+    // deckRef,
```
kylebarron: I'm not exactly sure how to pass through the `deckRef` to get a deck.gl reference. I figure I probably have to modify `DeckGLOverlay` somehow?

kylebarron:
Thanks!

> The lockfile is outdated, we should probably run npm i and update it before merging.

Oh I missed that; I made #927 to validate on CI; and then I can bump the lockfile in another PR.

PR #923, opened 2025-10-14
refactor: Refactor model initialization logic

kylebarron:
In ipywidgets, a "model" is the core of reactivity. In particular, a Lonboard `Map` is reactive, but every underlying layer is _also_ reactive, so that when a user changes any property on a layer, it automatically gets updated on the map. 

By extension, our [`Extension`](https://developmentseed.org/lonboard/latest/api/layer-extensions/) classes are also themselves models. These are separate Python classes that get serialized as their own reactive models on the JS side.

Soon we'll have more types of widget models that we want to be reactive, including views and basemap styles (#908).

Right now, model initialization is slightly hard-coded to handle layer initialization. This PR refactors the model initialization to be fully generic.

### DIFF
--- a/lonboard/_html_export.py
+++ b/lonboard/_html_export.py
@@ -1,0 +1,84 @@
+from __future__ import annotations
+
+from io import StringIO
+from typing import IO, TYPE_CHECKING, TextIO, overload
+
+from ipywidgets.embed import dependency_state, embed_minimal_html
+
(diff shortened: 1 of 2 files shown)


kylebarron:
Open to comments still, but this is a pretty straightforward refactor I think, so I'll merge so that I can use it in #935

Issue #681, opened 2024-10-09
Geocoding/search plugin/widget/control


prusswan:
**Is your feature request related to a problem? Please describe.**

Lonboard is great for visualizing large datasets, but sometimes I need to zero into a particular location so a search widget/control would be handy.

**Describe the solution you'd like**

Some means of building/integrating a search/geocoding control (that takes in textual input, turn it into location coordinates for the map to zoom/center to). User to provide JS event handling code for the respective search/geocoding service they prefer to use. But a sample can be provided for Nominatim/Mapbox etc

**Describe alternatives you've considered**

N/A. Might have to dig deeper into the widget internals if I had to do it myself

**Additional context**

I have some previous code done for folium (https://github.com/prusswan/folium/tree/multiple-and-generic-geocode-providers) - which is now able to support multiple geocoding services, although the custom geocode support (allowing users to define geocoding and event handling functions) was not incorporated due to maintenance reasons. (python-visualization/folium#1851)

prusswan:
> Geocoding needs some sort of service running in the background, so it's unclear if we want that as part of the native Lonboard JS app. 

yeah for this to work as part of the exported html, there needs to be a way for the html to access an external service (could be anything from places autocomplete or some other search endpoint), I guess I am just looking for a least intrusive way to get this to work.

> You could make your own [anywidget](https://github.com/manzt/anywidget)-based widget with a little bit of JavaScript, and then wire up an event to set lonboard state.

A pure-Python widget should be sufficient for using within Jupyter. But if this can be part of the html, this will be more useful to the users who just want the visualization without having to deal with Python or JS. Actually, I just found #203 which is probably the other control I would need as well. Do you see any potential changes for lonboard to make it easier for users to add such controls (e.g. write less widget code)?

kylebarron:
> A pure-Python widget should be sufficient for using within Jupyter. But if this can be part of the html, this will be more useful to the users who just want the visualization without having to deal with Python or JS.

Well, the ideal solution would be to have some sort of `lonboard-controls` library, that has its own controls that integrate with Lonboard via Python widget properties. Though this means the controls will be rendered next to the map instead of on the same HTML object.

prusswan:
![ol_autocomplete_anywidget](https://github.com/user-attachments/assets/d9343766-2013-4352-9748-e3948a2feb70)


[...]

PR #924, opened 2025-10-14
feat: Support for map controls

kylebarron:
### Change list

- Adds definitions for map "controls" (which deck.gl calls "widgets", but the name "widget" is too overloaded in the context of Jupyter, so I'd like to call them "controls")
- Support for `ScaleControl`, `NavigationControl`, `FullscreenControl`
- These controls are supported by both deck.gl and maplibre, and so implementations are defined for creating both the deck.gl and maplibre counterparts

### Example

```py
from lonboard import Map
from lonboard.controls import ScaleControl, FullscreenControl, NavigationControl

m = Map([], 
    basemap=MaplibreBasemap(mode="overlaid"),
    controls=[ScaleControl(), FullscreenControl(), NavigationControl()]
)
m
```

image

We should probably change the default positioning of controls, move the bbox selector, or make the bbox selector a control itself.

----

However, for deck.gl-driven rendering, I can only see the scale control:
image

Logged in https://github.com/developmentseed/lonboard/issues/990

### Todo list

- Figure out how to pass these react components as children of the `OverlayRenderer` or `DeckFirstRenderer` (depends on https://github.com/developmentseed/lonboard/pull/921)
- Use the refactored model initialization from https://github.com/developmentseed/lonboard/pull/923 to create the list of control instances on the deck side
- Integrate more cleanly into the Python API (an extension of https://github.com/developmentseed/lonboard/pull/908)

Closes https://github.com/developmentseed/lonboard/issues/842, relevant to https://github.com/developmentseed/lonboard/issues/681

### DIFF
--- a/src/toolbar.tsx
+++ b/src/toolbar.tsx
@@ -28,7 +28,7 @@ const Toolbar: React.FC = () => {
     <div
       style={{
         position: "absolute",
-        top: "20px",
+        bottom: "20px",
         right: "20px",
         zIndex: 1000,
         backgroundColor: "white",


kylebarron:
I moved the bounding box selection widget to the bottom right

image

Issue #524, opened 2024-05-14
Automated widget testing


kylebarron:
There's now [pytest-ipywidgets](https://pypi.org/project/pytest-ipywidgets/).

See https://github.com/jupyter-widgets/ipywidgets/issues/2934#issuecomment-2109499356. Note that this might still be quite difficult to implement because we need to emulate a GPU as well if we want to render data. I believe deck used some setup with xvfb to emulate the GPU in CI

kylebarron:
Fuller docs: https://solara.dev/documentation/advanced/howto/testing#testing-with-a-browser

PR #935, opened 2025-10-15
feat: Create richer `Basemap` class and deprecate `basemap_style` arg

kylebarron:
### Change list

- Add `basemap` parameter to `Map`. This now manages a class with more information than just the basemap style. In particular, this holds the render mode `"interleaved"`, `"overlaid"`, or `"reverse-controlled"` that was implemented on the JS side in https://github.com/developmentseed/lonboard/pull/921
- Deprecate `basemap_style` parameter to `Map`. This is superseded by the `style` parameter to `MaplibreBasemap`
- Rename `CartoBasemap` to `CartoStyle`
- Add `before_id` parameter to the core `Layer`. When passed and the basemap render mode is `"interleaved"`, 
- Pass `interleaved` prop down to MapboxOverlay
- Remove temporary `render_mode` from #921

**Rendering layer interleaved in maplibre layer stack:**

image

TODO:

- [ ] Validation for `beforeId`. If `beforeId` is passed to a layer, try to resolve the basemap style URL to a dict, and then validate that the string value of `beforeId` exists in that layer stack. Keep a cache of fetched styles so that you're not fetching them repeatedly.
- [ ] Warn if `beforeId` is set when the basemap style is not `"interleaved"`?
- [ ] (next PR): integration with https://github.com/geopandas/xyzservices (see #494)
- [ ] (possibly future PR): examples of rendering different basemap modes

For https://github.com/developmentseed/lonboard/issues/494


lonboard/basemap.py
```diff
@@ -42,4 +116,62 @@ class CartoBasemap(str, Enum):
...
     )
-    """A light, colored map style without labels."""
+    """A light, colored map style without labels.
+
+    ![](https://carto.com/help/images/building-maps/basemaps/voyager_no_labels.png)
+    """
+
+
+class MaplibreBasemap(BaseWidget):
+    """A MapLibre GL JS basemap."""
```
kylebarron, line 126: Improve this docstring

lonboard/basemap.py
```diff
@@ -42,4 +116,62 @@ class CartoBasemap(str, Enum):
...
+        self,
+        *,
+        mode: Literal[
+            "interleaved",
+            "overlaid",
+            "reverse-controlled",
+        ] = "reverse-controlled",
+        style: str | CartoStyle = CartoStyle.PositronNoLabels,
+    ) -> None:
+        """Create a MapLibre GL JS basemap."""
```
kylebarron, line 138: todo: improve this docstring

docs/api/basemap.md
```diff
@@ -1,3 +1,3 @@
 # lonboard.basemap
 
-::: lonboard.basemap.CartoBasemap
+::: lonboard.basemap.CartoStyle
```
kylebarron, line 3: Need to add `MaplibreBasemap` to the docs here

src/index.tsx
```diff
@@ -120,11 +118,40 @@ function App() {
...
+  const [basemapState, setBasemapState] = useState<MaplibreBasemapModel | null>(
+    null,
+  );
+
   // Fake state just to get react to re-render when a model callback is called
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [stateCounter, setStateCounter] = useState<Date>(new Date());
   const updateStateCallback = () => setStateCounter(new Date());
 
+  useEffect(() => {
```
kylebarron, line 134: todo: in the future, we should figure out a way to refactor some of this code out of the top-level `App`, because it's pretty ugly having all of this layer model, basemap model, view model, etc, state in the top-level body.

lonboard/traits.py
```diff
@@ -1078,7 +1078,6 @@ def __init__(
         **kwargs: Any,
     ) -> None:
         super().__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
kylebarron: It's more consistent to have `.tag(sync=True)` called where the trait is used, not inside its constructor.

lonboard/_map.py
```diff
@@ -203,17 +219,43 @@ def on_click(self, callback: Callable, *, remove: bool = False) -> None:
     - Default: `5`
     """
 
-    basemap_style = BasemapUrl(CartoBasemap.PositronNoLabels)
-    """
-    A URL to a MapLibre-compatible basemap style.
+    basemap: t.Instance[MaplibreBasemap | None] = t.Instance(
+        MaplibreBasemap,
+        allow_none=True,
+    ).tag(
```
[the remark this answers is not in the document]
kylebarron: `allow_none=True` allows `None`

lonboard/basemap.py
```diff
@@ -42,4 +116,62 @@ class CartoBasemap(str, Enum):
...
+    - **`"reverse-controlled"`**:
+
+        The reverse-controlled mode renders deck.gl above the MapLibre container and blocks any interaction to the base map.
+
+        If you need to have multiple views, you should use this option.
+
+    **Default**: `"reverse-controlled"`
+    """
+
+    style = BasemapUrl(CartoStyle.PositronNoLabels).tag(sync=True)
```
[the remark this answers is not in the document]
kylebarron: `BasemapUrl` is imported

vgeorge:
@kylebarron considering it's possible to pass `None` to the base map prop, shouldn't the Map component render a blank canvas? Right now it's still displaying the default basemap. Just wondering if that's something we’re addressing in this PR or planning to handle separately.

kylebarron:
Indeed I was planning to add that in the next PR, but I realized it's actually easier than I thought to add, so I can add it here

kylebarron:
Well, it "works" 
image
but now the red background I thought was always some sort of deck.gl rendering bug... is clear it's actually the deck.gl canvas background color...

kylebarron:
It turns out that the line 

```css
// globals.css
@tailwind utilities;
```

was causing the background to be red. And taking that out removes the red background.

I was thinking this was edited in #740, but actually it was added all the way back in #417. Perhaps there was a tailwind update that caused `@tailwind utilities` to make the background red?

image

This also removes the red "flash" that I've seen for a long time before maplibregl renders above the red background.

But removing that line is problematic because it breaks sidebar and bbox styling:

image
image

There must be another way to fix the red background

kylebarron:
The JS-only end-to-end testing is valuable, but I'd really love to have Python-driven testing of the widgets. We can discuss on https://github.com/developmentseed/lonboard/issues/524. Maybe we should test out pytest-ipywidgets (https://solara.dev/documentation/advanced/howto/testing#testing-with-a-browser)

PR #971, opened 2025-10-23
refactor: Manage views, layers, basemap in react hooks

kylebarron:
### Change list

- Refactor out verbose state handling of views, layers, basemap into react hooks
- Create `setSelectedBounds` callback from `useModelState("selected_bounds")`

Closes https://github.com/developmentseed/lonboard/issues/967

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -196,7 +196,7 @@ function App() {
     pickingRadius: pickingRadius,
     onClick: onMapClickHandler,
     onHover: onMapHoverHandler,
-    useDevicePixels: isDefined(useDevicePixels) ? useDevicePixels : true,
+    ...(isDefined(useDevicePixels) && { useDevicePixels }),
     onViewStateChange: (event) => {
       const { viewState } = event;
 


kylebarron:
Oh this is probably something like... the `deps` of the `useEffect` compare equality shallowly, but two functions always compare as not equal in JS, so passing in a function callback as a dep probably causes it to re-render every frame.

kylebarron:
> Oh this is probably something like... the `deps` of the `useEffect` compare equality shallowly, but two functions always compare as not equal in JS, so passing in a function callback as a dep probably causes it to re-render every frame.

That's exactly what it was 🙂

kylebarron:
I tested that selected bounds from the map were still getting propagated through to the Python session

kylebarron:
_whoops_ I didn't push the last commit fixing the behavior before merging the PR 🤦‍♂️

kylebarron:
Well actually it looks like Github is maybe struggling with git actions right now, so I clicked "git push" but it never pushed

kylebarron:
https://github.com/developmentseed/lonboard/pull/972 contains the last commit that was supposed to go in this branch before merging

Issue #955, opened 2025-10-20
DataFilterExtension seems to be broken on main (may be on my end though...)


ATL2001:
## Context

While working on the get_category_filter PR, I was getting some strange behavior that seemed to suddenly break what I was working on.  I decided to pull main down to my machine and see if possibly something else besides what I was doing somehow caused it.  When I run the code below (based on the data filter extension docs, with the colors and slider stripped out) the points from the scatter plot layer aren't rendered on the map.

<img width="1270" height="754" alt="Image" src="https://github.com/user-attachments/assets/2e534bde-520f-4eb7-86f7-da61e0f0314f" />

If I don't add the extension, it renders the points.

<img width="1270" height="754" alt="Image" src="https://github.com/user-attachments/assets/bf8a6b89-5964-4b25-85f1-380030704c48" />

If someone could try to recreate this and either confirm that it's not just me, or let me know that it's on my end only I'd really appreciate it.  I'm scratching my head right now.

## Environment

- OS: Windows 11
- Browser: opera
- Lonboard Version: Main

## Steps to reproduce the bug

```py
from pathlib import Path

import geopandas as gpd
import numpy as np
import pandas as pd
import shapely

import lonboard
from lonboard import Map, ScatterplotLayer
from lonboard.layer_extension import DataFilterExtension

url = "https://ookla-open-data.s3.us-west-2.amazonaws.com/parquet/performance/type=mobile/year=2019/quarter=1/2019-01-01_performance_mobile_tiles.parquet"
local_path = Path("data-filter-extension.parquet")
if local_path.exists():
    gdf = gpd.read_parquet(local_path)
else:
    columns = ["avg_d_kbps", "avg_u_kbps", "avg_lat_ms", "devices", "tile"]
    df = pd.read_parquet(url, columns=columns)

    tile_geometries = shapely.from_wkt(df["tile"])
    tile_centroids = shapely.centroid(tile_geometries)
    non_geom_columns = [col for col in columns if col != "tile"]
    gdf = gpd.GeoDataFrame(
        df[non_geom_columns],
        geometry=tile_centroids,
        crs="EPSG:4326",
    )
    gdf.to_parquet(local_path)

filter_extension = DataFilterExtension(filter_size=3)

filter_values = np.column_stack(
    [gdf["avg_d_kbps"], gdf["avg_u_kbps"], gdf["avg_lat_ms"]],
)
initial_filter_range = [
    [10_000, 50_000],
    [1000, 10_000],
[...]

PR #977, opened 2025-10-26
fix: Restore DataFilterExtension functionality with vectorized accessor input

kylebarron:
The existing code was no longer working after https://github.com/developmentseed/lonboard/pull/912 because we're passing Arrow _`Vector`s_ into the accessor props instead of `arrow.Data` objects.

This PR fixes that by selecting a specific record batch of data to pass into each layer.

In https://github.com/geoarrow/deck.gl-layers/issues/164 upstream, it should be easier to catch errors like this.

Closes https://github.com/developmentseed/lonboard/issues/955

### DIFF
--- a/src/model/base-layer.ts
+++ b/src/model/base-layer.ts
@@ -7,6 +7,7 @@ import type {
   PickingInfo,
 } from "@deck.gl/core";
 import type { WidgetModel } from "@jupyter-widgets/base";
+import { Vector } from "apache-arrow";
 
 import { BaseModel } from "./base.js";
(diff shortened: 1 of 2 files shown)


ATL2001 (approved):
YES👏 I didn't pull this branch down but I did a copy/paste to get the relevant parts into my branch (which I'll discard so as to not cause commit issues) and it's all working again!  

I figured it must have had something to do with the switch from the arrow table to the recordbatches, and now that I see what you're doing here, it makes a lot of sense why I wasn't able to find anywhere in deckgl-layers to do something like this.

PR #991, opened 2025-10-28
wip: Geocoder control

### LINKED ISSUE #681: Geocoding/search plugin/widget/control

kylebarron:
This is complex enough that it's not going to make it into this release

- The Maplibre geocoder implementation doesn't provide _any_ out of the box support for any provider
- It's hard to match up the deck.gl widget backends to the maplibre geocoder API

Closes #681

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@deck.gl/react": "^9.2.2",
         "@geoarrow/deck.gl-layers": "^0.4.0-beta.1",
         "@geoarrow/geoarrow-js": "^0.3.2",
+        "@maplibre/maplibre-gl-geocoder": "^1.9.1",
         "@nextui-org/react": "^2.4.8",
         "@xstate/react": "^6.0.0",
         "apache-arrow": "^21.1.0",
@@ -3065,6 +3066,25 @@
       "license": "ISC",
       "engines": {
         "node": ">=6.0.0"
+      }
+    },
+    "node_modules/@maplibre/maplibre-gl-geocoder": {
+      "version": "1.9.1",
+      "resolved": "https://registry.npmjs.org/@maplibre/maplibre-gl-geocoder/-/maplibre-gl-geocoder-1.9.1.tgz",
+      "integrity": "sha512-tLd93wQWwr9l/svNTYG25WT5Bo1KQMXwUl21Y0OOB8bV0TMhmbEzs1OUUBuWmo6Xn07lhgm5Y5GZvn7eMfmz1A==",
+      "license": "ISC",
+      "dependencies": {
+        "events": "^3.3.0",
+        "lodash.debounce": "^4.0.6",
+        "subtag": "^0.5.0",
+        "suggestions-list": "^0.0.2",
+        "xtend": "^4.0.1"
+      },
+      "engines": {
+        "node": ">=18"
+      },
+      "peerDependencies": {
+        "maplibre-gl": ">=4.0.0"
       }
     },
     "node_modules/@maplibre/maplibre-gl-style-spec": {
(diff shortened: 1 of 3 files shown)


prusswan:
Closest to out of the box would be their example which uses Nominatim: https://maplibre.org/maplibre-gl-js/docs/examples/geocode-with-nominatim/

kylebarron:
I would've ideally liked to provide a selection of a few geocoder providers that users could choose from, because the actual geocoding API calls should be able to integrate into each library's solution. I don't really want to add maplibre-only or deck-only support, because library that's rendering the map view should be just an implementation detail.

prusswan:
> I would've ideally liked to provide a selection of a few geocoder providers that users could choose from, because the actual geocoding API calls should be able to integrate into each library's solution. I don't really want to add maplibre-only or deck-only support, because library that's rendering the map view should be just an implementation detail.

If you have an idea of which providers you want to support (for a start), we can explore possible options. It is quite hard to support multiple geocoding providers in the long term, when most of them are commercial, while it is not the main priority of  library authors either. I don't normally use Nominatim as I would need to self-host to use it for autocomplete (due to their usage limit). 

Maplibre geocoder design is still useful as a general framework for mapping disparate geocoding inputs to outputs. Do you think it is possible to support one or two providers, and allow users to define their own? Through JS code etc

kylebarron:
Ideally I would choose the top N geocoding providers by usage (not sure what those are), including at least one free one, if possible.

I don't currently have a way in Lonboard for users to provide their own JS code. Going through Python for geocoding requests might be too slow

eKerney:
@kylebarron has anyone considered [geopy](https://github.com/geopy/geopy) integration into lonboard? 
Perhaps it could be an optional dependency.  They have quite a few providers, and several free options for small requests: Nominatim, Geocodio, OpenCage etc... [geopy.geocoders](https://github.com/geopy/geopy/tree/master/geopy/geocoders)    
Even dare I say... ESRI has a geolocation API that doesn't require a key for small numbers of requests.

prusswan:
> @kylebarron has anyone considered [geopy](https://github.com/geopy/geopy) integration into lonboard? Perhaps it could be an optional dependency. They have quite a few providers, and several free options for small requests: Nominatim, Geocodio, OpenCage etc... [geopy.geocoders](https://github.com/geopy/geopy/tree/master/geopy/geocoders) Even dare I say... ESRI has a geolocation API that doesn't require a key for small numbers of requests.

It's a good idea, but for my use-case (visualization + location search without backend dependency for live usage) I ended up making a custom widget that works in the exact way I want (e.g. don't use Nominatim with autocomplete to stay within their limits, a handful of commercial providers, custom geocode functions etc), basically what I originally requested. I'm guessing this is what most people end up doing

kylebarron:
Actually, going through Python does simplify things a lot and remove complexity from our side in Lonboard. I'll put up a new PR this afternoon

eKerney:
Cross this one off the list, nice work Kyle! ✔️

PR #1041, opened 2025-11-06
draft: Generic tiling systems

kylebarron:
image

https://github.com/user-attachments/assets/638d96da-9b8b-4e9d-9d39-e1d69d8ad09f

(this is actually not slowed down 😅 so we have some performance improvements to do!)

> This summary was drafted by claude and then I've gone through and edited it

This PR is a draft, trying to implement support for managing a `TileLayer` based on an arbitrary tile layout, which in this case happens to be the internal tile layout of a generic COG image in an arbitrary coordinate system.

The goal is that deck.gl will automatically fetch tiles as needed, based on the current viewport, from either internal overviews or the full resolution data. This will be generic enough that we can extend it to Zarr data later.

deck.gl exposes the [`Tileset2D`](https://deck.gl/docs/api-reference/geo-layers/tile-layer#tileset2d) class, which we can pass into a `TileLayer` to override how it fetches tiles. So we need to implement a subclass of `Tileset2D` with custom behavior.

The implementation tries to follow deck.gl's own traversal for OSM web mercator indexing as much as possible, and tries to infer where the important differences are between web mercator tiling and generic tiling.

### COG Metadata Extraction

- Extracts overview levels, tile dimensions, CRS information from GeoTIFF
- Computes affine geotransforms for each overview level
- Sets up proj4 transformation objects for coordinate reprojection

### Tile Traversal

- Implements quadtree-based tile traversal starting from coarsest overview
- [Frustum culling](https://en.wikipedia.org/wiki/Viewing_frustum) to only render visible tiles
- [In progress but not working]: Distance-based Level of Detail (LOD) selection following deck.gl's OSM tile implementation
- Handles COGs with multiple tiles at the coarsest level (unlike OSM which always has 1 tile at z=0)

### Coordinate Transformations

The implementation handles a complete transformation pipeline:

1. **Pixel coordinates → Image CRS** using affine geotransform
2. **Image CRS → Web Mercator (EPSG:3857)** using proj4
3. **Web Mercator → deck.gl world space [0, 512]** for frustum culling

### Tile Bounds Calculation

Tile bounds are computed using the geotransform at 9 reference points across the tile surface:

1. Sample 9 points for each tile: corners, center, midpoints of each side
2. Transform each point from pixel space to image CRS with geotransform
3. Reproject from image CRS to Web Mercator (EPSG:3857) with proj4
4. Convert to deck.gl ["common space"](https://deck.gl/docs/developer-guide/coordinate-systems#common-space) (where the web mercator axes are normalized to `[0, 512]`)
5. Create an `Oriented Bounding Box` from common space points (which then should work with deck.gl's frustum culling implementation)

This approach handles:
- Arbitrary CRS (not just Web Mercator)
- Potential rotation/skew in imagery
- Non-rectangular tiles in common space

### Working ✅

- COG metadata extraction with proper geotransform derivation
- Coordinate transformations (Image CRS → Web Mercator → World Space)
- Frustum culling (tiles render in correct geographic location)
- Tile bounds visualization via PathLayer
- Basic tile rendering infrastructure

### In Progress ⚠️

- **LOD selection:** Currently always selects finest resolution tiles regardless of zoom level
- Need to tune distance-based LOD calculation to properly account for COG scale factors
- The algorithm needs adjustment to select coarser tiles when zoomed out

### Not Yet Implemented ❌

- Tile data loading from COG (reading pixel data from GeoTIFF)
- Actual raster tile rendering (separate work stream)

## References

- [deck.gl TileLayer Documentation](https://deck.gl/docs/api-reference/geo-layers/tile-layer)
- deck.gl [Tileset2D docs](https://deck.gl/docs/api-reference/geo-layers/tile-layer#tileset2d)
- [deck.gl tile traversal implementation for OSM indexing](https://github.com/visgl/deck.gl/blob/9bfe769183e1671c2d20d7c42bffe165a9dd109a/modules/geo-layers/src/tileset-2d/tile-2d-traversal.ts)
- [Initial tile traversal PR](https://github.com/visgl/deck.gl/pull/4593) (👋 hi me from 5 years ago)
- Python `affine`: https://www.perrygeo.com/python-affine-transforms.html

### DIFF
--- a/cog-tile-testing.ipynb
+++ b/cog-tile-testing.ipynb
@@ -1,0 +1,166 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "id": "66285659-d958-4be6-ae62-b3b6efdad127",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from lonboard import Map\n",
+    "from lonboard.experimental._surface import COGTileLayer"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "id": "49ea9e5f-ee08-4924-b968-2e18c058c876",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "url = \"https://ds-wheels.s3.us-east-1.amazonaws.com/m_4007307_sw_18_060_20220803.tif\"\n",
+    "layer = COGTileLayer(data=url)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b890fe4a-6291-4dd6-b62a-20581e22404d",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "view_state = {'longitude': -73.20972420038129,\n",
+    " 'latitude': 40.90307721701271,\n",
+    " 'zoom': 10.98785761041711}"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "id": "21aa3c2f-e8d3-4a92-acf1-2c1af8451de7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m = Map(layer, view_state=view_state)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 11,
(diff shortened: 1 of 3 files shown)


kylebarron:
This was initial prototyping that became part of https://github.com/developmentseed/deck.gl-raster

PR #1042, opened 2025-11-07
refactor: migrate from XState to Zustand for state management

### LINKED ISSUE #916: Refactor XState to simpler state management
The current XState implementation is overkill for our simple UI interactions. 

**Current state:** 3-state machine with 6 event types for basic interactions:
- Bbox selection workflow (start → end → complete)
- Map click/hover events  
- Toolbar button states
- Side panel visibility
- Feature highlighting

**Proposal:** Evaluate simpler alternatives (Zustand, XState store, or React state) that better match the UI complexity. Create a proof-of-concept branch, migrate incrementally, and use existing Playwright tests to verify all interactions work.

**Goal:** Lower barrier to entry for JS developers while maintaining all functionality.

vgeorge:
## Summary
- Replace XState machine with Zustand store for simplified state management
- Maintain all existing functionality including bbox selection and feature highlighting
- Add comprehensive test coverage for new state management

Fixes #916

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -9,6 +9,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: PR Conventional Commit Validation
-        uses: ytanikin/pr-conventional-commits@1.4.2
+        uses: ytanikin/pr-conventional-commits@1.5.1
         with:
           task_types: '["feat","fix","docs","test","ci","refactor","perf","chore","revert"]'


kylebarron (commented):
Thanks for the PR!

We currently have two different places for state management:

- backbone models (via jupyter widgets) for any state coming in from Python
- xstate -> zustand for any "within JS application/UI state"

Do you think this split makes sense? Should we try to unify all state in a single place?

vgeorge:
@kylebarron thanks for the review. I think we should keep them separate. The parts handled by Zustand don’t need to be exposed, this avoids unnecessary back-and-forth serialization for UI-only state. I added some changes to better document this approach. This is ready for another review.

vgeorge:
@kylebarron I added more test coverage to include the tooltip on feature hover. I won't be adding more changes to this until your feedback.

src/state/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * State management for Lonboard.
+ *
+ * Two types of state:
+ * - Client-only: UI state that never syncs with Python (Zustand)
+ * - Python-synced: State that bidirectionally syncs with Python (Backbone)
```
kylebarron: Maybe just add a tiny note that `store.ts` holds the client-only state


```suggestion
 * - Client-only: UI state that never syncs with Python (Zustand). See `store.ts`
 * - Python-synced: State that bidirectionally syncs with Python (Backbone). See `python-sync.ts`
```

src/state/python-sync.ts
kylebarron, line 1: We have a lot of state in different places. Essentially the entire folder of `src/model` (https://github.com/developmentseed/lonboard/tree/main/src/model) is just definitions for state to be synced between Python and JS.

Do you intend for Zustand to be used with any of that? Would Zustand help at all beyond the current events that backbone propagates?

src/index.tsx
```diff
@@ -261,9 +333,7 @@ function App() {
 
 const WrappedApp = () => (
   <NextUIProvider>
-    <MachineProvider>
```
kylebarron, line 264: What does this do? Was `MachineProvider` part of `XState`?

src/index.tsx
```diff
@@ -27,38 +29,83 @@ import {
...
+        getLineColor: [0, 0, 0, 130],
+        stroked: true,
+        getLineWidth: 2,
+        lineWidthUnits: "pixels",
+      };
+      if (isDrawingBbox) {
+        bboxProps.getFillColor = [255, 255, 0, 120];
+        bboxProps.getLineColor = [211, 211, 38, 200];
+        bboxProps.getLineWidth = 2;
+      }
```
kylebarron, line 104: These can always be set on `bboxProps`. The canonical way to toggle visibility is with the layer's [`visible` prop](https://deck.gl/docs/api-reference/core/layer#visible). So we can set `visible: false` in general and then 
```ts
if (isDrawingBbox) {
  bboxProps.visible = true;
}
```

kylebarron:
Sorry for the slow review! It's been a super busy month

src/index.tsx
vgeorge: Yes, this is not necessary anymore.

src/state/python-sync.ts
vgeorge: @kylebarron I think Zustand might help us avoid the `updateStateCallback` pattern, but I’m not fully sure yet. My suggestion is to keep things as they are for now, and if we notice performance issues around this area, we can revisit and explore a small bridge.

src/index.tsx
vgeorge: Ok! I'll refactor on a separate PR.

vgeorge: I gave this a bit more thought. I don’t think using the `visible` prop helps here, since deck.gl would still need to manage a layer that shouldn’t exist in the first place. Happy to revisit if you see other advantages.

PR #1055, opened 2025-12-04
fix: Fix map.set_view_state does not set pitch and bearing

### LINKED ISSUE #1052: [BUG] map.set_view_state does not set pitch and bearing
After updating to 0.13, map.set_view_state() calls no longer update the bearing or pitch. I looked for the breaking commit and found that it was 2ac3d52e8af21f81ea365228c066a897df52ffc0

## Environment

- OS: Windows 11
- Browser: Edge
- Lonboard Version: 0.13.0

## Steps to reproduce the bug
from lonboard import viz

```python
m = viz(data=[])
m.set_view_state(longitude=-80, latitude=50, zoom=2, pitch=45, bearing=90)
m
```

This should produce a map that is 1) at significant pitch and 2) rotated 90 degrees. Instead, it produces

Image

jwardbond:
Previously, changing pitch and bearing using set_view_state or by clicking and dragging the map did not forward updates to Map.view_state. Added tests to verify.

Closes #1052


src/util.ts
jwardbond, line 1: initially wrote an isMapView() check, as suggested https://github.com/developmentseed/lonboard/issues/1052#issuecomment-3604412745 but turns out view was always "undefined" when I created my maps with `m = viz([])`

tests/test_map.py
jwardbond, line 1: I don't know the peculiarities of jupyter widgets, but these tests always passed, even with pitch and bearing specified.

tests/ui/test_mapview_interaction.py
jwardbond, line 1: I added some tests. This was fun to play around with. 

Oddly, updating pitch/bearing with `set_view_state` and  then calling `view_state` in the same cell correctly reflected the changes (in the text output, not on the map) but if `view_state` was instead called in a subsequent cell, pitch and bearing were not updated. This is my best attempt at testing that behavior, and the manual pitch/bearing adjustment behaviour.

kylebarron:
Thank you!

PR #1060, opened 2025-12-08
gt scratch work

### LINKED ISSUE #1: Separate into multiple widgets/layers?

kylebarron:
## What I am changing

-

## How I did it

-

## How you can test it

-

## Related Issues

-


kylebarron:
This was prototyped before https://github.com/developmentseed/deck.gl-raster became a standalone library.

PR #1131, opened 2026-03-15
fix: Make sure BaseLayer.highlight_color has .tag(sync=True)

### LINKED ISSUE #1072: [BUG] Setting a custom highlight color gets ignored
## Context

When I set the highlight color for a layer, that is the color that should get used when hovering over an object.

Despite explicitly setting a highlight color, the default highlight color gets used.

## Environment

- OS: Windows and Ubuntu
- Browser: Edge and Firefox
- Lonboard Version: 0.12.1 and master

## Steps to reproduce the bug

I generated a map using the following code snippet and hovered over the polygon:

```python
from shapely import Polygon
import geopandas as gpd
from lonboard import Map, PolygonLayer

coords = ((-107., 35.), (-107., 39.), (-105., 39.), (-105., 35.))
polygon = Polygon(coords)

gdf = gpd.GeoDataFrame({"name": ["my_shape"], "geometry": [polygon]})
layer = PolygonLayer.from_geopandas(gdf, get_fill_color="#FF000033", pickable=True, auto_highlight=True)
layer.highlight_color = [0, 255, 0]  # Should be bright green
map_ = Map(layer)
map_
```

I can attempt to take on fixing this one, but I'm still learning how ipywidget works.

shawnboltz:
Closes #1072.

### DIFF
--- a/lonboard/layer/_base.py
+++ b/lonboard/layer/_base.py
@@ -213,8 +213,6 @@ class BaseLayer(BaseWidget):
     - Default: `False`
     """
 
-    # Note that the default_value of VariableLengthTuple doesn't accept an empty list
-    #  or None
     highlight_color = VariableLengthTuple(
         t.Int(),
         default_value=(0, 0, 128, 128),


lonboard/layer/_base.py
```diff
@@ -213,12 +213,14 @@ def _add_extension_traits(self, extensions: Sequence[BaseExtension]) -> None:
...
+    # Note that the default_value of VariableLengthTuple doesn't accept an empty list
+    #  or None
     highlight_color = VariableLengthTuple(
         t.Int(),
-        default_value=None,
+        default_value=[0, 0, 128, 128],
         minlen=3,
         maxlen=4,
-    )
+    ).tag(sync=True)
```
kylebarron, line 221: Thank you! 

Now that I think of it, perhaps we should have wrappers of traitlets types so that we don't have to remember to call `.tag(sync=True)`

lonboard/layer/_base.py
```diff
@@ -213,12 +213,14 @@ def _add_extension_traits(self, extensions: Sequence[BaseExtension]) -> None:
     - Default: `False`
     """
 
+    # Note that the default_value of VariableLengthTuple doesn't accept an empty list
+    #  or None
     highlight_color = VariableLengthTuple(
         t.Int(),
-        default_value=None,
+        default_value=[0, 0, 128, 128],
```
kylebarron: I think we want the default to stay as `None`

https://github.com/developmentseed/lonboard/blob/0a3b86310f88ff9869291d136b6501c62c11ce81/lonboard/layer/__init__.py#L3-L8

shawnboltz: Ok. It was failing when the default value was None because it was converting to an empty list and the VariableLengthTuple wouldn't allow it. I can dig into that more.

PR #1137, opened 2026-03-20
feat: raster layer async geotiff integration

kylebarron:
https://github.com/user-attachments/assets/38f350de-8a23-49fa-9c0a-79cc585350da

### Change list

- Manage creation of `fetch_tile` closure inside of `from_geotiff`
- Propagate request cancellation from `AbortSignal` through to Python, so it can cancel asyncio tasks if needed
- New `TileMatrixSetTrait` for validating and serializing TMS definitions
- New `ProjectionTrait` for validating and serializing projections
- Rename `cog` optional feature to `raster`
- Bump deck.gl-raster dependency to 0.4.0
- Always use deck.gl dependency from local `node_modules` (useful if linking to local directory of deck.gl-raster)

### TODO: 

- [ ] prepare examples for release
- [ ] Update example to show a process for debugging the tile visualization
- [ ] Add `raster.reshape_as_image` helper
- [ ] Pass over API docs
- [ ] Write release post

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,0 @@
-# Stately Inspector is enabled when this is set to true
-XSTATE_INSPECT=false
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -51,10 +51,10 @@ npm run build:watch
 
 ### Environment Variables
 
-To use custom environment variables, copy the example environment file:
+To use custom environment variables, you can create a file `.env`:
 
-```sh
-cp .env.example .env
+```bash
+VARIABLE="setting"
 ```
 
 This file contains the list of environment variables for the JavaScript component, and the build task will use them when available.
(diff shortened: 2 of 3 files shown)


lonboard/layer/_raster.py
```diff
@@ -285,46 +293,54 @@ def render(tile: Buffer | None) -> EncodedImage | None:
...
+            _bounds=Bbox(*wgs84_bounds),
+            _center=wgs84_center,
+            **kwargs,
+        )  # type: ignore[call-arg]
 
     _layer_type = t.Unicode("raster")
 
-    # TODO: Restore TMS generic tile traversal. For now, for simplicity, we're only rendering standard web mercator tiles.
-    # _tms = Dict()
+    tile_matrix_set: TileMatrixSet = TileMatrixSetTrait(allow_none=True).tag(sync=True)  # type: ignore
```
kylebarron: Perhaps we should make this private for now, so that it can only be set through the constructors.

PR #1148, opened 2026-03-25
chore: Update Python lockfile

kylebarron:
Tests failing due to https://github.com/widgetti/solara/issues/1143


kylebarron:
Looks like the referenced issue has been closed so I can try to return to this

kylebarron:
I think this was updated in a subsequent PR

PR #1162, opened 2026-04-03
ci: add uv package manager to dependabot

lhoupert:


kylebarron (approved):
thanks

kylebarron:
for my knowledge: what's the usefulness of this for lonboard? lonboard is a library, not an application, so uv.lock is only checked in for development convenience and reliability. but it doesn't really "matter" what version is in my personal development lock file, because that doesn't change at all what version the end user installs. 

And I don't want to restrict the range of versions beyond the minimal version that is required for lonboard to functionally work. 

@lhoupert

lhoupert:
Very good point @kylebarron ! I thought initially it was used in the ci too, but looking at the CI more closely, I see test.yml uses uv sync without --locked, so the lock file isn't strictly enforced. 

The main benefit here is mainly keeping the lock file fresh for local dev convenience. 

But it got me thinking, would you be open to switching CI to --locked? That way the dependabot uv PRs would keep CI reproducible . If this fails, you know your code or a specific dependabot bump broke it. 

Or we could keep the existing test.yml that is good in catching up compatibility with newer steps, and create another test workflow that use --locked ?

kylebarron:
Yeah, I'd be happy to and probably should use `--locked` in CI, but that still seems irrelevant to using dependabot for these dependencies.



> The main benefit here is mainly keeping the lock file fresh for local dev convenience.

So what's the benefit to using dependabot? If the environment is only used for local testing, there are virtually no security issues compared to having a bundle that is distributed as an application

kylebarron:
As far as I know, uv will use the lockfile automatically unless it's inconsistent with the `pyproject.toml`. Since I always check in the lockfile, `--locked` wouldn't change anything in practice other than being a check on my work.

PR #1163, opened 2026-04-03
fix: overture-geoparquet example

### LINKED ISSUE #1027: Overture example doesn't work because overture deleted their data
The overture-geoparquet example doesn't work anymore

```
> aws s3 ls s3://overturemaps-us-west-2/release/
                           PRE 2025-09-24.0/
                           PRE 2025-10-22.0/
```

From the overture slack:
Image

That is so dumb. I'll let someone else update the notebook (it should be possible to list the releases in the bucket and choose one)

matyasrada:
making the release version dynamic by listing the S3 bucket with obstore.list_with_delimiter instead of hardcoding a release date

Closes #1027

### DIFF
--- a/examples/overture-geoparquet.ipynb
+++ b/examples/overture-geoparquet.ipynb
@@ -3,7 +3,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "57e4678a",
+   "id": "0",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -28,7 +28,7 @@
     }
    },
    "cell_type": "markdown",
-   "id": "250a25da-9547-4a0a-b709-8659e0669545",
+   "id": "1",
    "metadata": {},
    "source": [
     "# Spatially-Partitioned GeoParquet\n",
@@ -44,7 +44,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "665edcb2-b546-435e-ba77-7ea651ab6ae6",
+   "id": "2",
    "metadata": {},
    "source": [
     "## Dependencies\n",
@@ -60,7 +60,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "62c1bef5-071e-4cf0-98e4-2e70171c3006",
+   "id": "3",
    "metadata": {},
    "source": [
     "## Imports\n"
@@ -69,7 +69,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "9ae52a41",
+   "id": "4",
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


examples/overture-geoparquet.ipynb
```diff
@@ -85,37 +86,80 @@
...
    "source": [
     "store = S3Store.from_url(\n",
-    "    \"s3://overturemaps-us-west-2/release/2025-06-25.0/\",\n",
+    "    \"s3://overturemaps-us-west-2\",\n",
     "    region=\"us-west-2\",\n",
     "    skip_signature=True,\n",
-    ")"
+    ")\n",
+    "\n",
+    "result = obs.list_with_delimiter(store, prefix=\"release\")\n",
```
kylebarron: Could you use `store.list_with_delimiter` instead of `obs.list_with_delimiter(store, ...)`

examples/overture-geoparquet.ipynb
```diff
@@ -85,37 +86,80 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "store = S3Store.from_url(                                                                                                                                                  \n",
+    "      f\"s3://overturemaps-us-west-2/{latest}/\",                                                                                                                              \n",
+    "      region=\"us-west-2\",                                                                                                                                                    \n",
+    "      skip_signature=True,                                                                                                                                                   \n",
+    "  )"
+   ]
+  },
```
kylebarron: We don't need to create another store instance. We can just prepend `{latest}` onto the paths passed below.

examples/overture-geoparquet.ipynb
kylebarron: Also, can you put this after the commented-out code block that uses the Azure store? Because finding the latest release should be the same for both backends.

kylebarron (commented):
Thanks for the PR! Just a couple comments

matyasrada:
Hello Kyle! 

Thanks for the review!

I updated a few things:

- removed importing obs
- using store.list_with_delimiter() instead of obs.list_with_delimiter
- removed one store and prepending {latest} to paths instead
- moved release listing after the Azure block

one problem is that it seems to me Azure doesn't support list_with_delimiter so it can't find the latest release dynamically. i might be wrong. also with this logic if we found it through s3 first as a workaround we still prepended down the cells and so we double-referenced the prefix.

kylebarron:
> one problem is that it seems to me Azure doesn't support list_with_delimiter so it can't find the latest release dynamically. i might be wrong

You just needed to keep the `release` string on the Azure bucket. I fixed the notebook in case you want to try yourself

kylebarron:
Thanks!

PR #1179, opened 2026-04-23
wip: Start brainstorming zarr integration

### LINKED ISSUE #1: Separate into multiple widgets/layers?

kylebarron:
Designs a Python-proxied Zarr tile pipeline via RasterLayer.from_zarr and RasterLayer.from_xarray, reusing the existing GeoTIFF model. Async per-tile fetching via zarr-python 3 directly or xarray.load_async. JS-side raster model gains a generic TilesetDescriptor trait; TileMatrixSet becomes one descriptor shape. Scope narrowly to Python-rendered PNGs; WebGL rendering, animation, and RasterTileLayer refactor are deferred.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

## What I am changing

-

## How I did it

-

## How you can test it

-

## Related Issues

-


kylebarron:
Does xarray define any chunking natively? Lonboard needs to know the internal chunk size of the input dataset, or the user needs to pass them in.

vincentsarago:
@kylebarron looking at the brainstorming document I see the API for `from_zarr` or `from_xarray` to accept `arrays`. Just for your information, when the zarr has multiscale, the user should be able to pass a zarr.group (xarray.dataset/datatree).

PR #1201, opened 2026-06-07
adds a one-way fly_to_command traitlet enabling flyTo from JS

batpad:
Currently, if one wants the map to `flyTo` a location, one has to send a message through Jupyter comms. Thus, it is not possible to update the map view when `lonboard` is rendered in static contexts.

@kylebarron I know this is something you're not keen on supporting :( and I can see that the design is a lot nicer when going through comms. But it would be really nice to have some way to update the viewport in a frontend-only environment.

This adds a `fly_to_command` traitlet that is attached to the shared `model`, so can be set in frontend JS. This implementation should work for me, but am not sure of the naming, etc. Just putting this up as a draft PR and then we can discuss.

### DIFF
--- a/lonboard/_map.py
+++ b/lonboard/_map.py
@@ -201,6 +201,15 @@ class Map(BaseAnyWidget):
     Indicates if a click handler has been registered.
     """
 
+    _fly_to_command = traitlets.Dict(allow_none=True, default_value=None).tag(sync=True)
+    """Imperative camera command consumed by the frontend (one-way: control -> map).
+
+    Keys mirror ``fly_to()``: ``type``, ``longitude``, ``latitude``, ``zoom``, ``pitch``,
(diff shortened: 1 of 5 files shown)


kylebarron:
Ok so if I understand:

From a client side only app you're not able to go through Jupyter Comms? Are you able to mock it some way so that the Lonboard JS side doesn't have to change?

The bigger question I think is that if we proceed on this sort of client-side work, do we need a client-side public API separate from the Python API?

batpad:
@kylebarron your idea of just mocking the comms on the frontend when a kernel isn't present seems like it should work, so we don't need to have these two divergent paths in `lonboard`.

Closing this, pretty sure the work-around is fine, and maybe in fact useful as a generic way to mock comms communication for static widgets. Thanks!

PR #1205, opened 2026-06-12
fix: Support CRS-parameterized GEOMETRY type in DuckDB 1.5

kylebarron:
Written by Claude:

----

### Problem

DuckDB 1.5's spatial extension parameterizes the `GEOMETRY` type with a CRS (e.g. `GEOMETRY('EPSG:4326')`). Lonboard detects geometry columns in `from_duckdb` by exact string comparison against `"GEOMETRY"`, which no longer matches, so `viz()` and `Layer.from_duckdb` fail with `ValueError: No geometry column found` for any relation read via `ST_Read` under DuckDB ≥ 1.5. (Found while attempting a lockfile upgrade: 11 tests in `tests/test_duckdb.py` fail with duckdb 1.5.3.)

### Fix

Match geometry columns by the base type name with type parameters stripped, which behaves identically on DuckDB 1.4. Once detected, the existing `ST_AsWKB` conversion path works unchanged on both versions.

Tests are made version-agnostic:
- `GEOMETRY` type assertions accept the parameterized form.
- `ST_AsWKB` returns plain `BLOB` (not `WKB_BLOB`) in DuckDB 1.5, so the WKB tests cast its result with `::WKB_BLOB` to keep exercising lonboard's `WKB_BLOB` path.
- DuckDB 1.5 has no cast from CRS-parameterized `GEOMETRY` to `POINT_2D` (`ConversionException` at execution), so the `POINT_2D` tests cast through plain `GEOMETRY` first (a no-op on 1.4).

`uv.lock` is intentionally untouched; the dependency bump can land separately once this merges.

### Verification

- duckdb 1.4.4 (locked), full suite: `185 passed, 1 skipped`
- duckdb 1.5.3 (`uv run --with duckdb==1.5.3`), `tests/test_duckdb.py`: `14 passed, 1 skipped` (was `11 failed, 3 passed, 1 skipped` before the fix)

### Follow-up

DuckDB 1.5 exposes the CRS via the type parameter and attaches `geoarrow.wkb` extension metadata (with PROJJSON CRS) to its Arrow export; lonboard could propagate that CRS instead of warning that none exists. Kept out of scope here to stay minimal.

### DIFF
--- a/lonboard/_geoarrow/_duckdb.py
+++ b/lonboard/_geoarrow/_duckdb.py
@@ -35,13 +35,13 @@ def _base_type_name(duckdb_type: duckdb.typing.DuckDBPyType) -> str:
     """Return the name of a DuckDB type without any type parameters.
 
     As of DuckDB 1.5, the spatial extension's GEOMETRY type is parameterized
-    with a CRS, e.g. ``GEOMETRY('EPSG:4326')``, so an exact string comparison
-    against ``"GEOMETRY"`` no longer matches.
+    with a CRS, e.g. `GEOMETRY('EPSG:4326')`, so an exact string comparison
+    against `"GEOMETRY"` no longer matches.
 
-    String parsing is the only way to do this: ``DuckDBPyType`` exposes only
-    ``id`` and ``children``, where ``id`` omits the type parameters (and is
-    ``'blob'`` for both GEOMETRY and WKB_BLOB on DuckDB 1.4, so it can't be
-    used for detection either) and ``children`` raises for non-nested types.
+    String parsing is the only way to do this: `DuckDBPyType` exposes only
+    `id` and `children`, where `id` omits the type parameters (and is
+    `'blob'` for both GEOMETRY and WKB_BLOB on DuckDB 1.4, so it can't be
+    used for detection either) and `children` raises for non-nested types.
     """
     return str(duckdb_type).split("(", 1)[0]
 
--- a/lonboard/_geoarrow/_duckdb.py
+++ b/lonboard/_geoarrow/_duckdb.py
@@ -31,13 +31,23 @@ DUCKDB_SPATIAL_TYPES = {
 }
 
 
+def _base_type_name(duckdb_type: duckdb.typing.DuckDBPyType) -> str:
+    """Return the name of a DuckDB type without any type parameters.
+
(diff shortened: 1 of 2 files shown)


lonboard/_geoarrow/_duckdb.py
```diff
@@ -31,13 +31,23 @@
...
 
 
+def _base_type_name(duckdb_type: duckdb.typing.DuckDBPyType) -> str:
+    """Return the name of a DuckDB type without any type parameters.
+
+    As of DuckDB 1.5, the spatial extension's GEOMETRY type is parameterized
+    with a CRS, e.g. ``GEOMETRY('EPSG:4326')``, so an exact string comparison
+    against ``"GEOMETRY"`` no longer matches.
+    """
+    return str(duckdb_type).split("(", 1)[0]
```
kylebarron, line 46: Is it only the `GEOMETRY` type that can be parameterized? Or can say `POINT_2D` also be parametrized?

kylebarron, line 46: Also is there a way to get the parametrized parts from the `DuckDBPyType` directly instead of casting to a string and using string parsing?

kylebarron: From Claude:

> Only `GEOMETRY` — verified empirically on duckdb 1.5.3: every other spatial type rejects parameters with `Binder Error: Type 'POINT_2D' does not take any type parameters` (same for `LINESTRING_2D`, `POLYGON_2D`, `BOX_2D`, `WKB_BLOB`). So the stripping here is only ever exercised for `GEOMETRY`, and is a no-op for the rest.

kylebarron: From Claude:

> No — `DuckDBPyType` exposes exactly two attributes: `id` and `children`. For a CRS-parameterized geometry type, `id` is `'geometry'` (parameter not included) and `children` raises `Invalid Input Error: This type is not nested so it doesn't have children`. So string parsing is the only access to the parameter from the Python API.
> 
> `id` also can't replace the string match for detection: duckdb 1.4's GEOMETRY has `id == 'blob'`, and `WKB_BLOB` has `id == 'blob'` in both versions — indistinguishable from a plain `BLOB` column.

PR #1206, opened 2026-06-12
fix: Support h3 4.5.0 

kylebarron:
Compute default H3 viewport per resolution for h3 4.5 compatibility

Written by Claude:
----

### Problem

h3 4.5.0's `cells_to_geo` (via `cells_to_h3shape`) raises `H3ResMismatchError` when passed cells of mixed resolutions. `default_h3_viewport` passes the full cell sample in a single call, so constructing an `H3HexagonLayer` from mixed-resolution cells crashes during default-viewport computation under h3 ≥ 4.5. (Found while attempting a lockfile upgrade; any user with mixed-resolution cells would hit this.)

### Fix

Group the sampled cells by resolution (using lonboard's existing vectorized `get_resolution` bit-mask helper, no extra h3 calls) and call `cells_to_geo` once per resolution — at most 16 calls — then derive the bbox/centroid from the combined boundary coordinates. Behavior is unchanged for single-resolution inputs and works under both h3 4.4 and 4.5. This preserves the boundary-based bbox semantics; a centroid-only approach would understate the bbox for coarse cells.

Adds `test_mixed_resolution_default_viewport`, which builds a layer from one cell at each resolution 0–15 and asserts a real finite viewport is computed. Verified red under h3 4.5.0 before the fix (`H3ResMismatchError`), green under both versions after.

`uv.lock` is intentionally untouched; the dependency bump can land separately once this merges.

### Verification

- h3 4.4.2 (locked), full suite: `186 passed, 1 skipped`
- h3 4.5.0 (`uv run --with h3==4.5.0`), `tests/test_h3/`: `34 passed`

### DIFF
--- a/lonboard/layer/_h3.py
+++ b/lonboard/layer/_h3.py
@@ -7,6 +7,7 @@ import numpy as np
 
 import lonboard.traits as t
 from lonboard._geoarrow.ops import Bbox, WeightedCentroid
+from lonboard._h3._validate_h3_cell import get_resolution
 from lonboard._utils import auto_downcast as _auto_downcast
 from lonboard.layer._polygon import PolygonLayer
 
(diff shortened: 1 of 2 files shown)


kylebarron (commented):
lgtm

PR #1210, opened 2026-06-12
feat: Read CRS from DuckDB 1.5 GEOMETRY columns

kylebarron:
### Summary

Lonboard previously assumed all DuckDB geometry data was EPSG:4326. DuckDB 1.5's spatial extension encodes a CRS on the `GEOMETRY` type (e.g. `GEOMETRY('EPSG:3857')`) and exports such columns over Arrow as spec-compliant GeoArrow WKB with the full PROJJSON CRS in the extension metadata. This PR uses that CRS automatically: data in a non-4326 CRS is now reprojected correctly instead of being rendered as if it were lon/lat.

Follow-up to #1205 (see the design spec in `dev-docs/specs/2026-06-12-duckdb-crs-support-design.md`).

### Implementation

- `_from_geometry` now consumes DuckDB's native Arrow export directly when the exported field carries the `geoarrow.wkb` extension name (DuckDB ≥ 1.5) — no more `ST_AsWKB` round-trip, which was where the CRS was lost. Detection is runtime feature detection on the field metadata, not a version check; DuckDB 1.4 exports plain binary and falls back to the existing `ST_AsWKB` path unchanged.
- New `_reconcile_crs` helper implements the `crs=` parameter semantics against the column CRS:
  - column CRS present + matching `crs=` → `UserWarning` that passing `crs` is no longer needed with DuckDB ≥ 1.5 when using a `GEOMETRY` type with a CRS encoded
  - column CRS present + conflicting `crs=` → `ValueError` naming both CRS
  - column CRS absent → `crs=` applied exactly as before, no warning
- `crs=` is soft-deprecated in the `from_duckdb` docstring: it remains necessary for `WKB_BLOB`/2D columns (which cannot carry a CRS — verified they reject type parameters), DuckDB < 1.5, and `GEOMETRY` columns without an encoded CRS. The `viz()` docstring info box is updated to match.
- Side effect of dropping the SQL interpolation: non-alphanumeric geometry column names now work on DuckDB ≥ 1.5 (`test_sanitize_column_name` updated to pin both behaviors).

### Test plan

- [x] New tests: CRS read from `GEOMETRY('EPSG:3857')` and auto-reprojected (asserts the map centers at the true lon/lat); redundant-match warning; mismatch error; CRS-less column + `crs=` unchanged
- [x] Full suite under locked duckdb 1.5.3: `195 passed, 1 skipped`
- [x] `tests/test_duckdb.py` under duckdb 1.4.4 (`uv run --with duckdb==1.4.4`): `14 passed, 5 skipped` — legacy `ST_AsWKB` path regression-tested
- [x] `ruff check` clean; `uv.lock`/`pyproject.toml` untouched


lonboard/_geoarrow/_duckdb.py
```diff
@@ -114,6 +115,79 @@ def _from_geometry(
...
+    table: Table,
+    *,
+    geom_col_idx: int,
+    crs: str | pyproj.CRS | None,
+) -> Table:
+    """Apply the `crs` parameter to a GeoArrow table exported from DuckDB.
+
+    - Column has a CRS and `crs` matches it: warn that `crs` is unnecessary.
+    - Column has a CRS and `crs` conflicts with it: raise `ValueError`.
+    - Column has no CRS: attach `crs` (if provided) like before DuckDB 1.5.
```
kylebarron: These should be "if column ..."

lonboard/_geoarrow/_duckdb.py
```diff
@@ -114,6 +115,79 @@ def _from_geometry(
...
+    """
+    import pyproj
```
kylebarron: imports should be at top level unless they're conditional, but pyproj is a non-conditional dependency

lonboard/_geoarrow/_duckdb.py
```diff
@@ -114,6 +115,79 @@ def _from_geometry(
...
+
+    geom_field = table.schema.field(geom_col_idx)
+    field_metadata = geom_field.metadata or {}
+    ext_meta_bytes = field_metadata.get(b"ARROW:extension:metadata")
+    ext_meta = json.loads(ext_meta_bytes) if ext_meta_bytes else {}
+    column_crs = ext_meta.get("crs")
```
kylebarron: Do we not have a generic helper for parsing a CRS from geoarrow field metadata?

lonboard/_geoarrow/_duckdb.py
kylebarron: Reworded to "If the column has a CRS and ..." in eb4c434.

🤖 Written by [Claude Code](https://claude.com/claude-code)

lonboard/_geoarrow/_duckdb.py
kylebarron: Moved `import pyproj` to the top level in eb4c434, and removed the same pre-existing inline import from `_make_geoarrow_field_metadata` while there. `duckdb` stays under `TYPE_CHECKING` since it's an optional dependency.

🤖 Written by [Claude Code](https://claude.com/claude-code)

lonboard/_geoarrow/_duckdb.py
kylebarron: There is one — `lonboard._geoarrow.crs.get_field_crs`. Switched to it in eb4c434, which also drops the manual `json.loads` and `CRS.from_json_dict` and gains handling for the non-projjson `crs_type` variants (`wkt2:2019`, `authority_code`) that the hand-rolled version ignored.

🤖 Written by [Claude Code](https://claude.com/claude-code)

lonboard/_geoarrow/_duckdb.py
```diff
@@ -3,8 +3,10 @@
 import json
 import re
 from typing import TYPE_CHECKING
+from warnings import warn
 
 import numpy as np
+import pyproj
```
kylebarron: use `from pyproj import CRS` canonically instead of import pyproj

kylebarron: Switched to `from pyproj import CRS` throughout the file (annotations and call sites) to match `lonboard/_geoarrow/crs.py`. Done in 2027f33.

🤖 Written by [Claude Code](https://claude.com/claude-code)

kylebarron (commented):
lgtm

Issue #1024, opened 2025-10-31
Linked maps example broken


kylebarron:
Need to clean up view state handling

keewis:
just to mention, linking view states is much better using `ipywidgets.jslink` (if that's the only thing we do), because that doesn't need to roundtrip to and from the kernel and thus doesn't have a lag between the linked maps.

For example:
```python
map1 = ...
map2 = ...

ipywidgets.jslink((map1, "view_state"), (map2, "view_state"))
```

PR #1214, opened 2026-06-21
fix: map.set_view_state and linked maps example

### LINKED ISSUE #1062: [BUG] set_view_state no longer moves previously rendered map
## Context

Calling `set_view_state` on a map with a new lat/lon isn't moving a previously rendered map.  I've poked around in the code a bit, but it's not jumping out to me where the issue is.  I suspect it's the same underlying problem which is causing #1024 and is what [jwardbond is seeing](https://github.com/developmentseed/lonboard/pull/1055#discussion_r2587092364).

**Describe what happened:**

Image

## Environment

- OS: Windows 11
- Browser: opera
- Lonboard Version: 0.13.0

## Steps to reproduce the bug
cell 1
```py
from lonboard import viz
m = viz(data=[])
m.set_view_state(longitude=-62, latitude=45, zoom=5) # moves map `m` to Nova Scotia
m
```
cell2
```py
m.set_view_state(longitude=-90, latitude=45, zoom=5) # should move map `m` to Wisconsin
# If this cell was run really quickly after cell 1 (as in running a notebook with restart kernel
# and run all cells), map `m` in cell 1 will have moved to wisconsin
# but if I let the map in cell 1 render then run this cell and look at the map displayed in cell 1
# it will still be centered in Nova Scotia
# 
# No matter what, if we display the map again here, it is in Wisconsin...
m
```

ATL2001:
This PR fixes the `set_view_state` function from the python side by calling `fly_to` with `duration=0`, and updates the linked maps example to make it functional again.  Admittedly, this is not the ideal solution, of fixing it on the JS side, but it does get the set_view_state function and linked maps example working.  I thought I'd push it up as a fix until someone with the necessary JS/TS/React skills to fix it properly has time. If you'd rather hold off for the ideal fix, I won't take it personally 😄. 

While I was in there I noticed that we weren't checking that the pitch and bearing inputs of `fly_to` were numeric, so I added that in there as well.

fixes #1062 #1024

### DIFF
--- a/examples/linked-maps.ipynb
+++ b/examples/linked-maps.ipynb
@@ -70,7 +70,7 @@
     "import traitlets\n",
     "\n",
     "from lonboard import Map\n",
-    "from lonboard.basemap import CartoBasemap\n",
+    "from lonboard.basemap import CartoStyle, MaplibreBasemap\n",
     "from lonboard.view_state import MapViewState"
    ]
   },
@@ -96,7 +96,7 @@
     "## Create postitron map focused on the arch\n",
     "positron_map = Map(\n",
     "    layers=[],\n",
-    "    basemap_style=CartoBasemap.Positron,\n",
+    "    basemap=MaplibreBasemap(style=CartoStyle.Positron),\n",
     "    view_state={\n",
     "        \"longitude\": -90.1849,\n",
     "        \"latitude\": 38.6245,\n",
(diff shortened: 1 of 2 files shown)


ATL2001:
didn't realize there were tests in here for this. looks like I'm hitting a race condition.  before the map wasn't getting updated, but the properties were, now the map is getting updated but the properties aren't updated immediately😞.. i really don't want to introduces a delay into the function... perhaps this won't work

ATL2001:
I got around the race condition, but using fly_to won't work with the GlobeViewState or FirstPersonViewState, I'm going to scrap this.

Issue #1232, opened 2026-07-20
Make an example for Geohash layer


aryanxk02:
With reference to https://github.com/developmentseed/lonboard/issues/311 and https://github.com/developmentseed/lonboard/issues/138, I would like to contribute few examples for small dataset groups around themes such as Logistics and Transportation, Climate Hazard Risks – wildfires, deforestation, etc and Oceanography. I recently tried out lonboard to map out airports from all over the world. Check the image below for reference.

<img width="2766" height="1530" alt="Image" src="https://github.com/user-attachments/assets/1952d2fc-140c-47e8-8b00-9f018fb3c3db" />

I think this would be a great place for new users to understand and explore lonboard's vast scope to map out various datasets. 

Kindly let me know what you think @emmalu, @kylebarron. I am ready to work on this enthusiastically.

kylebarron:
Hi @aryanxk02, thanks for making this issue.

In terms of the examples here, I'm mostly thinking of having examples be tailored to the _how_. How to use various different layer types, how to string things together. 

The _what_ and the _why_ I think are better left to end users. I don't think it's worth maintaining examples for how to use specific datasets, for example. End users will know their datasets much better than I ever could.

We also have links to third party examples. If you want to make and host any of your own examples, we can add a link in the [Third Party showcase section](https://developmentseed.org/lonboard/latest/examples/#third-party-showcase).

aryanxk02:
Hi @kylebarron , thanks for the reply! I would like to know what could a potential example for Geohash Layer look like...do we implement different parameters associated with the function or is there any specific thing/workflow you would like to highlight and create a new example?

kylebarron:
I think a potential Geohash layer example could be super simple; just take some geohash data and plot it on a map. The idea would be more to explain what Geohash is and how to use it. Unless you know of a data source that's already in Geohash and which would be useful to show in an example

aryanxk02:
Sounds good. If no one else is already working on it, I'd be happy to take this. I'll build an example demonstrating the usage of Geohash Layer and open a PR. Later we can iterate on it during the review process and refine it as per your suggestions. Please let me know.

kylebarron:
I don't think anyone else is working on a geohash example

aryanxk02:
Okay. I will make a PR soon and reference this issue there.

PR #1242, opened 2026-08-06
feat: Add an example for Geohash Layer

aryanxk02:
This PR adds an example demonstrating the Geohash Layer. Reference: https://github.com/developmentseed/lonboard/issues/1232


examples/geohash-layer.ipynb
```diff
@@ -0,0 +1,289 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
```
kylebarron, line 4: Similar to other examples like https://github.com/developmentseed/lonboard/blob/main/examples/data-filter-extension.ipynb, could you create the example notebook with [`juv`](https://github.com/manzt/juv), so that dependencies are automatically managed with uv?

Assuming you have `uv` installed already, you can use something like `uvx juv init [name]` to create the notebook, `uvx juv add [name] [dependency]` to add dependencies, and `uvx juv run [name]` to run the notebook.

Then you can copy in the markdown instructional cell:

Image

examples/geohash-layer.ipynb
```diff
@@ -0,0 +1,289 @@
+{
```
kylebarron, line 1: CI is failing on some formatting and linting on this notebook

examples/geohash-layer.ipynb
```diff
@@ -0,0 +1,289 @@
...
+    "jupyter": {
+     "source_hidden": true
+    }
+   },
+   "outputs": [],
+   "source": [
+    "# /// script\n",
+    "# requires-python = \">=3.12\"\n",
+    "# dependencies = [\n",
+    "#     \"geohash2\",\n",
```
kylebarron: geohash2 is 10 years old: https://github.com/dbarthe/geohash/

Can we use another more modern geohash library like https://pypi.org/project/pygeohash/?

examples/geohash-layer.ipynb
```diff
@@ -0,0 +1,289 @@
...
+    "\n",
+    "requests_df = pd.DataFrame(response.json())\n",
+    "if requests_df.empty:\n",
+    "    raise RuntimeError(\"The NYC 311 query returned no requests.\")\n",
+    "\n",
+    "requests_df = requests_df.astype({\"latitude\": \"float64\", \"longitude\": \"float64\"})\n",
+    "requests_df[\"geohash\"] = [\n",
+    "    geohash2.encode(latitude, longitude, precision=6)\n",
+    "    for latitude, longitude in zip(requests_df[\"latitude\"], requests_df[\"longitude\"], strict=True)\n",
+    "]\n",
```
kylebarron, line 212: It would be _nice_ if the geohash library provided a vectorized API, so that we could nudge people towards a faster approach. But this is fine.

kylebarron (commented):
Overall looking good, thank you

aryanxk02:
> Overall looking good, thank you

Thank you for the review. Let me know if any changes are needed.

kylebarron:
Thanks

